PR #283, opened 2022-11-25
refactor: use provided name as default value for created entity

eoaksnes:

* Use name from request

* The dm-core package and `NewEntityButton` are taking the name as input, so this should be the value used when creating the entity


KristianKjerstad:
ready for review?

PR #291, opened 2022-11-29
fix: include package json file with package meta info when exporting

### LINKED ISSUE #286: Include `_meta_` in export for root package and single document

KristianKjerstad:
 include package json file with package meta info when exporting.
handle 3 cases when exporting:
1) single file
2) root package
3) non root package

non root pacakge is not yet supported. will add functionality for this later

support export feature
closes #286

### DIFF
--- a/src/features/export/use_cases/export_use_case.py
+++ b/src/features/export/use_cases/export_use_case.py
@@ -3,28 +3,35 @@ import tempfile
 import zipfile
 
 from authentication.models import User
+from common.exceptions import ApplicationException
 from domain_classes.tree_node import Node
+from enums import SIMOS
 from services.document_service import DocumentService
 from storage.repositories.zip import ZipFileClient
 
 
-def create_zip_export(document_service: DocumentService, absolute_document_ref: str) -> str:
-    """Create a temproary folder on the host that contains a zip file with."""
+def create_zip_export(document_service: DocumentService, absolute_document_ref: str, user: User) -> str:
+    """Create a temporary folder on the host that contains a zip file with."""
     tmpdir = tempfile.mkdtemp()
     archive_path = os.path.join(tmpdir, "temp_zip_archive.zip")
 
     data_source_id, document_path = absolute_document_ref.split("/", 1)
     document: Node = document_service.get_by_path(absolute_document_ref)
-    # TODO add meta information to the document. Single document,
-    # root package and non root packages needs to be handled in different ways
-    with zipfile.ZipFile(archive_path, mode="w", compression=zipfile.ZIP_DEFLATED, compresslevel=5) as zip_file:
-        # Save the selected node, using custom ZipFile repository
-        document_service.save(document, data_source_id, ZipFileClient(zip_file), update_uncontained=True)
-    return archive_path
+    if document.entity["type"] == SIMOS.PACKAGE.value and document.entity["isRoot"]:
+        # TODO handle root package
+        raise ApplicationException(message="Create zip export is only supported for a single document (not a package)")
+    elif document.entity["type"] == SIMOS.PACKAGE.value and not document.entity["isRoot"]:
+        # TODO handle non root pacakge
+        raise ApplicationException(message="Create zip export is only supported for a single document (not a package)")
+    else:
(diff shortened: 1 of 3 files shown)


src/storage/repositories/zip.py
```diff
@@ -24,8 +25,19 @@ def update(self, entity: dict, storage_recipe=None, **kwargs):
     def get(self, uid: str):
         return "Not implemented on ZipFile repository!"
 
-    def add(self, uid: str, document: dict):
-        return "Not implemented on ZipFile repository!"
+    def add(self, document: dict, path: str, filename: str = None):
+        """Add the provided document as a  json file to the zip file"""
+        if path[-1] == "/":
+            raise ValidationException(message=f"When adding file to Zip, path ({path}) should not end with a '/'")
```
soofstad, line 31: can we not just strip leading and trailing "/" here?

src/storage/repositories/zip.py
```diff
@@ -24,8 +25,19 @@ def update(self, entity: dict, storage_recipe=None, **kwargs):
...
+        json_data = json.dumps(document)
+        binary_data = json_data.encode()
+
+        if filename:
+            write_path = f"{path}/{filename}.json"
+        else:
+            write_path = f"{path}/{document['name']}.json"
```
soofstad, line 38: Perhaps the caller should handle this, so this function always uses the filename?
Also, what happens if the document does not have a ["name"]?

src/features/export/use_cases/export_use_case.py
```diff
@@ -3,28 +3,73 @@
...
+    (since a package is exported as a folder in the zip file).
+    Therefore, we must add a package.json file inside the package folder to store meta information.
+    """
+    meta_documents = get_package_structure_from_node(document_node, {})
+    for path, document in meta_documents.items():
+        storage_client.add(document=document, path=path, filename="package")
+
+
+def create_zip_export(document_service: DocumentService, absolute_document_ref: str, user: User) -> str:
+    """Create a temporary folder on the host that contains a zip file with."""
```
soofstad, line 46:
```suggestion
    """Create a temporary folder on the host that contains a zip file"""
```

src/features/export/use_cases/export_use_case.py
```diff
@@ -3,28 +3,73 @@
...
+    document_node: Node = document_service.get_by_path(absolute_document_ref)
+    if document_node.entity["type"] == SIMOS.PACKAGE.value and document_node.entity["isRoot"]:
+        with zipfile.ZipFile(archive_path, mode="w", compression=zipfile.ZIP_DEFLATED, compresslevel=5) as zip_file:
+            # Save the selected node, using custom ZipFile repository
+            storage_client = ZipFileClient(zip_file)
+            document_service.save(document_node, data_source_id, storage_client, update_uncontained=True)
+            store_package_meta_as_files(document_node=document_node, storage_client=storage_client)
+
+        return archive_path
+    elif document_node.entity["type"] == SIMOS.PACKAGE.value and not document_node.entity["isRoot"]:
```
soofstad, line 60: no need for elif and else here. As each block ether returns or raises and exception

src/features/export/use_cases/export_use_case.py
```diff
@@ -3,28 +3,73 @@
...
-        # Save the selected node, using custom ZipFile repository
-        document_service.save(document, data_source_id, ZipFileClient(zip_file), update_uncontained=True)
-    return archive_path
+    document_node: Node = document_service.get_by_path(absolute_document_ref)
+    if document_node.entity["type"] == SIMOS.PACKAGE.value and document_node.entity["isRoot"]:
+        with zipfile.ZipFile(archive_path, mode="w", compression=zipfile.ZIP_DEFLATED, compresslevel=5) as zip_file:
+            # Save the selected node, using custom ZipFile repository
+            storage_client = ZipFileClient(zip_file)
+            document_service.save(document_node, data_source_id, storage_client, update_uncontained=True)
+            store_package_meta_as_files(document_node=document_node, storage_client=storage_client)
```
soofstad, line 57: Feel like this should not be needed. That the save() function and the ZipFileClient should be able to handle it...

soofstad (changes requested):

netr0m (approved):

soofstad (approved):

PR #297, opened 2022-12-07
build: enable mypy

soofstad:
Enable mypy and fix 198 errors. Some of these were bugs waiting to happend.
Also:
 - Bump python version
 - Bump dependency verions
 - Move flak8 pre-commit hook to not use local repo


KristianKjerstad (commented):
nice!

KristianKjerstad (approved):

PR #316, opened 2022-12-19
Fix/translate path to alias in export

### LINKED ISSUE #310: feat(dmss): use dependencies alias to translate types to use alias prefix
Convert from path to alias to, e.g. "dmss://system/SIMOS/Package" to  "CORE:Package"

KristianKjerstad:
in export, translate path to alias, for example "dmss://system/SIMOS/Package" is translated to "CORE:Package" if the the document's meta has an alias that matches adress = "system/SIMOS" and protocol = "dmss"
part of export functionality
closes #310

### DIFF
--- a/src/common/utils/replace_reference_with_alias.py
+++ b/src/common/utils/replace_reference_with_alias.py
@@ -2,17 +2,12 @@ from common.exceptions import ApplicationException
 from domain_classes.dependency import Dependency
 
 
-def is_alias(string_to_check: str, dependencies: list[Dependency]) -> bool:
-    """Check if a given string is an alias or not.
-
-    string_to_check is an alias on the form <alias>:<path> or a document reference on the format <protocol>://<datasource>/<path>
+def has_dependency_alias(reference: str, dependencies: list[Dependency]) -> bool:
+    """Check if a given reference string on the format <protocol>://<datasource>/<path> can be
+    substituted with an alias on the format <alias>:<path>.
     """
     for dependency in dependencies:
-        if (
-            string_to_check.startswith(f"{dependency.alias}:")
-            and dependency.protocol not in string_to_check
-            and dependency.address not in string_to_check
-        ):
+        if reference.startswith(dependency.get_absolute_reference()):
             return True
     return False
 
@@ -33,19 +28,21 @@ def replace_reference_with_alias(reference: str, dependencies: list[Dependency])
 
 
 def replace_absolute_references_in_entity_with_alias(entity: dict, dependencies: list[Dependency]):
-    """Replace path in an entity with alias defined in a list of dependencies.
+    """Replace references in an entity with alias defined in a list of dependencies.
 
     (for example, replace "dmss://system/SIMOS/Package" with "CORE:Package".)
     """
     attributes_to_update = ("type", "attributeType", "_blueprintPath_")  # These keys may contain a reference
-    EXTENDS = "extends"  # extends is a special attribute in an entity that contains a list of referencences
+    EXTENDS = "extends"  # Extends is a special attribute in an entity that contains a list of references
     for attribute in entity:
+        if type(entity[attribute]) == dict:
+            replace_absolute_references_in_entity_with_alias(entity[attribute], dependencies)
         if type(entity[attribute]) == list and attribute != EXTENDS:
             for new_entity in entity[attribute]:
                 replace_absolute_references_in_entity_with_alias(new_entity, dependencies)
-        if attribute in attributes_to_update:
+        if attribute in attributes_to_update and has_dependency_alias(entity[attribute], dependencies):
             entity[attribute] = replace_reference_with_alias(entity[attribute], dependencies)
         elif attribute == EXTENDS:
             for index, reference in enumerate(entity[attribute]):
-                if not is_alias(reference, dependencies):
-                    entity[attribute][index] = replace_reference_with_alias(entity[attribute][index], dependencies)
+                if has_dependency_alias(reference, dependencies):
+                    entity[attribute][index] = replace_reference_with_alias(reference, dependencies)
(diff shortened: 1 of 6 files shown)


src/common/utils/replace_reference_with_alias.py
```diff
@@ -0,0 +1,36 @@
+from domain_classes.dependency import Dependency
+
+
+def is_alias(path: str, dependencies: list[Dependency]):
+    NUM_SPLITS = 1
+    path_start = path.split(":", NUM_SPLITS)[0]
```
soofstad, line 6: This does not look right. Paths should never be aliases. Only absolute paths.
We need to check if we have an alias for an absolute path.

soofstad (changes requested):

src/common/utils/replace_reference_with_alias.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def replace_absolute_references_with_alias(entity: dict, dependencies: list[Dependency]):
+    """Replace path in an entity with alias defined in a list of dependencies.
+
+    (for example, replace "dmss://system/SIMOS/Package" with "CORE:Package".)
+    """
+    attribute_to_update = ("type", "attributeType", "extends", "_blueprintPath_")  # These keys may contain a reference
+    for attribute in entity:
+        if attribute == "attributes" or attribute == "content":
```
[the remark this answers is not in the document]
KristianKjerstad: Good point! Will update

src/common/utils/replace_reference_with_alias.py
KristianKjerstad: ah, yes. 
The function parameter is badly named. 

maybe something like this is better?
image

src/common/utils/replace_reference_with_alias.py
```diff
@@ -0,0 +1,36 @@
...
+            # "attributes" and "content" contain list of entities.
+            for new_entity in entity[attribute]:
+                replace_absolute_references_with_alias(new_entity, dependencies)
+        elif attribute in attribute_to_update:
+            for dependency in dependencies:
+                if attribute == "extends":
+                    # "extends" attribute contains a list of absolute references. Each list item is treated separately.
+                    for index, absolute_reference in enumerate(entity[attribute]):
+                        if not is_alias(entity[attribute][index], dependencies):
```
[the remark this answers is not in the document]
KristianKjerstad: The goal here is to handle the "extends" list separately, since it can contain a list of references: 
`    "extends": ["dmss://system/SIMOS/NamedEntity", "dmss://DemoApplicationDataSource/models/CarPackage/Vehicle"]`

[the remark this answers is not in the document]
soofstad: Allright good :+1:

src/common/utils/replace_reference_with_alias.py
soofstad: I dont understand why you need to check if it's an alias at all. We know it never will be any aliases in DMSS.
What makes sense to me is that we need a check to see if a "path" `has_dependency_alias()`. And if it does, replace the path with the alias.

src/common/utils/replace_reference_with_alias.py
```diff
@@ -0,0 +1,51 @@
...
+    """Replace path in an entity with alias defined in a list of dependencies.
+
+    (for example, replace "dmss://system/SIMOS/Package" with "CORE:Package".)
+    """
+    attributes_to_update = ("type", "attributeType", "_blueprintPath_")  # These keys may contain a reference
+    EXTENDS = "extends"  # extends is a special attribute in an entity that contains a list of referencences
+    for attribute in entity:
+        if type(entity[attribute]) == list and attribute != EXTENDS:
+            for new_entity in entity[attribute]:
+                replace_absolute_references_in_entity_with_alias(new_entity, dependencies)
```
soofstad, line 45: Looks like new_entity here could be a single string value.
And that would cause an exception in the call to this function.

src/common/utils/replace_reference_with_alias.py
```diff
@@ -0,0 +1,51 @@
...
+
+def replace_absolute_references_in_entity_with_alias(entity: dict, dependencies: list[Dependency]):
+    """Replace path in an entity with alias defined in a list of dependencies.
+
+    (for example, replace "dmss://system/SIMOS/Package" with "CORE:Package".)
+    """
+    attributes_to_update = ("type", "attributeType", "_blueprintPath_")  # These keys may contain a reference
+    EXTENDS = "extends"  # extends is a special attribute in an entity that contains a list of referencences
+    for attribute in entity:
+        if type(entity[attribute]) == list and attribute != EXTENDS:
```
soofstad, line 43: Does not look like we dig down into nested dicts here.
Should propably add that to the tests.

soofstad (changes requested):

src/common/utils/replace_reference_with_alias.py
KristianKjerstad: what is the best way to handle this? Add an extra check such that we only call the recursive function if type(new_entity) == object?

src/common/utils/replace_reference_with_alias.py
KristianKjerstad: ah, you are right.

It is no longer needed after refactoring the function. 
(before it was needed because of the structure of the loops - looping through dependencies before looping through strings in the extends lists)

src/common/utils/replace_reference_with_alias.py
soofstad: yeah think so.
Could take some inspiration from https://github.com/equinor/dm-cli/blob/main/dm_cli/import_package.py#L17.
It's not perfect, but this one will also be a bit simpler

KristianKjerstad:
todo return object instead of pass by ref

KristianKjerstad:
@soofstad  ready for a new review - tried to simplify the logic in replace_absolute_references_in_entity_with_alias() and updated replace_reference_with_alias_if_possible() to find the closest alias match instead of the first.

soofstad (changes requested):
Much better! :D

soofstad (approved):
:+1:

PR #318, opened 2022-12-21
fix: change type of version in Dependency.json to string

### LINKED ISSUE #317: fix(dmss): type for version in Dependency.json should be string

KristianKjerstad:
closes #317


soofstad (approved):
good catch!

PR #321, opened 2022-12-22
refactor: tree node serializer to separate file, and no longer a class

### LINKED ISSUE #313: Refactor(dmss): Move `DictImporter` / `to_dict` from `tree_node.py` (perhaps make it a simple function?)

soofstad:
- Move the logic from "DictImporter" and "DictExporter" to simple functions in a separate file
- Remove the Node.to_dict() functions  - To avoid circular dependencies and simpler to have ONE way to serialize Nodes

- More maintainable code base

closes #313

### DIFF
--- a/src/services/document_service.py
+++ b/src/services/document_service.py
@@ -12,6 +12,11 @@ from common.exceptions import (
     MissingPrivilegeException,
     NotFoundException,
 )
+from common.tree_node_serializer import (
+    tree_node_from_dict,
+    tree_node_to_dict,
+    tree_node_to_ref_dict,
+)
 from common.utils.build_complex_search import build_mongo_query
 from common.utils.delete_documents import delete_document
 from common.utils.get_blueprint import get_blueprint_provider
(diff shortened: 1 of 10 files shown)


src/services/document_service.py
```diff
@@ -173,9 +178,10 @@ def get_node_by_uid(self, data_source_id: str, document_uid: str, depth: int = 9
         complete_document = get_complete_sys_document(
             document_uid, self.repository_provider(data_source_id, self.user), depth
         )
-        return Node.from_dict(
+        return tree_node_from_dict(
             complete_document,
             complete_document.get("_id"),
+            key="",
```
KristianKjerstad, line 184: looks like key is set to "" in the majority of the tree_node_from_dict() function calls.

Maybe better to make key have a default value None or "" in tree_node_from_dict() and override it necessary?

KristianKjerstad (approved):

src/services/document_service.py
soofstad: fixed!
Did the same for UID

PR #322, opened 2023-01-02
Fix/update package blueprint and bug in remove_by_path

### LINKED ISSUE #319: fix: bug with storage_contained check in Node class
node.storage_contained gives different results from before https://github.com/equinor/data-modelling-storage-service/pull/309 was merged
-> this causes export zip functionality to fail

KristianKjerstad:
update package blueprint - let content list have contained = false as default. Also, add test for storage contained.

Also fixed a bug in remove_by_path - since node_id and uid for a node can be different, the wrong value was sent to the remove_by_child_id() function
bug with storage contained and remove_by_path
closes #319


src/services/document_service.py
```diff
@@ -370,7 +370,14 @@ def remove_by_path(self, data_source_id: str, directory: str):
             parent_uid = get_document_uid_by_path(f"{'/'.join(directory.split('/')[0:-1])}", data_source)
             child_uid = get_document_uid_by_path(directory, data_source)
             parent_node = self.get_node_by_uid(data_source_id, parent_uid)
-            parent_node.children[0].remove_by_child_id(child_uid)  # The first child of a directory is always 'content'
+            child_node_id = None
+
+            for child in parent_node.children[0].children:
+                if child.uid == child_uid:
+                    child_node_id = child.node_id
```
soofstad, line 377: For performance and readability, these kind of "find first match" should be written with the "next()" function

soofstad (approved):

PR #325, opened 2023-01-03
Perf/caching on lookups

### LINKED ISSUE #311: Feat(dmss): Re-implement caching on lookups
Caching was temporarily removed in #309 , but should be re-added once the feature has been completed

soofstad:
- Add caching for recipe lookup
- Add a "fetch lookup" endpoint. Useful for inspecting created lookups

Increase performance

closes #311


src/features/lookup_table/use_cases/get_lookup_table.py
```diff
@@ -0,0 +1,8 @@
+from authentication.access_control import DEFAULT_ACL, access_control
+from authentication.models import AccessLevel, User
+from storage.internal.lookup_tables import get_lookup
+
+
+def get_lookup_table_use_case(name: str, user: User) -> dict:
```
KristianKjerstad, line 6: from the name alone I don't really understand what value "name" should have.

in get_lookup() from lookup_tables.py it looks like an uid should be passed in. 
But from the endpoint /application/{application} it looks like an application name is passed in.

Maybe rename the parameter?

src/features/lookup_table/lookup_table_feature.py
```diff
@@ -26,3 +28,17 @@ def create_lookup(recipe_package: str, application: str, user: User = Depends(au
...
+
+
+@router.get(
+    "/application/{application}",
+    operation_id="get_lookup",
+    response_model=Lookup,
+    responses={**responses},
+)
+@create_response(JSONResponse)
+def get_lookup(application: str, user: User = Depends(auth_w_jwt_or_pat)):
```
KristianKjerstad, line 40: not sure what value "application" should have.
name or uid?

(Also see other comment)

README.md
```diff
@@ -72,20 +73,26 @@ poetry install
...
 ```
 
 Run unit tests:
 
-```
+```bash
 docker-compose run --rm dmss pytest api
 ```
 
+Creating DMSS lookup:
```
KristianKjerstad, line 92:
```suggestion
Creating DMSS lookup with [dm-cli](https://github.com/equinor/dm-cli):
```

src/features/lookup_table/use_cases/get_lookup_table.py
soofstad: renamed to lookup_id :+1:

src/features/lookup_table/lookup_table_feature.py
soofstad: There are no UUID anywhere related to lookups. Only "name" which also is used as the database key/_id

soofstad: I want to hide the "_id" logic from API users. Therefor it's only refered to as "application" in the endpoints.
But internally in the code, it's also acts as an ID.

KristianKjerstad (approved):
good!

PR #333, opened 2023-01-10
fix: update UiRecipe blueprint

KristianKjerstad:
- fix type in UiRecipe.json
various bug fixes

### DIFF
--- a/src/common/utils/create_entity.py
+++ b/src/common/utils/create_entity.py
@@ -22,7 +22,7 @@ class InvalidDefaultValue(CreateEntityException):
 
 
 class CreateEntity:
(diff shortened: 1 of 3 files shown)


KristianKjerstad:
bug with object type

image

soofstad (changes requested):
This has been discussed before.
"name" should not have a special treatment in "CreateEntity".
If there is a bug now, it should be fixed in some other way

KristianKjerstad:
> This has been discussed before. "name" should not have a special treatment in "CreateEntity". If there is a bug now, it should be fixed in some other way

fixed here instead https://github.com/equinor/dm-core-packages/pull/44/files

soofstad (approved):

PR #340, opened 2023-01-30
fix: be able to update list of objects

eoaksnes:

* Fixes so that it is possible to update list of objects

* Makes it possible to update list of objects

### DIFF
--- a/src/services/document_service.py
+++ b/src/services/document_service.py
@@ -163,13 +163,17 @@ class DocumentService:
         if node.type == SIMOS.BLOB.value:
             node.entity = self.save_blob_data(node, repository)
 
-        node.set_uid()  # Ensure the node has a _id
         ref_dict = tree_node_to_ref_dict(node)
 
-        entity_has_all_required_attributes(ref_dict, node.blueprint.get_required_attributes())
+        if type(node) is Node:
+            entity_has_all_required_attributes(ref_dict, node.blueprint.get_required_attributes())
 
         # If the node is not contained, and has data, save it!
         if not node.storage_contained and ref_dict:
+            # To ensure the node has a _id
+            if node.uid is None:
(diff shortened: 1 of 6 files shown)


.tool-versions
```diff
@@ -0,0 +1 @@
+python 3.11.1
```
soofstad, line 1: This file added by mistake?
Or; why is it needed?

src/services/document_service.py
```diff
@@ -163,13 +163,17 @@ def save(
...
 
-        entity_has_all_required_attributes(ref_dict, node.blueprint.get_required_attributes())
+        if type(node) is Node:
+            entity_has_all_required_attributes(ref_dict, node.blueprint.get_required_attributes())
 
         # If the node is not contained, and has data, save it!
         if not node.storage_contained and ref_dict:
+            # To ensure the node has a _id
+            if node.uid is None:
+                raise ApplicationException(f"The document with name `{node.name}` is missing uid")
```
soofstad, line 175: This should never happend. I gues the Node() instantiation should make sure the node has a correct ".uid"

soofstad (changes requested):

src/services/document_service.py
eoaksnes: I also think this should be handled correctly in the tree. But since I removed the old `node.set_uid()`, then I'm thinking that we might for a period check that every node had an uid and raise an Exception if not. Not sure why the node.set_uid() was there in the first place, so if someone knows why it was there and if it is still needed we can remove this check.

.tool-versions
eoaksnes: remove it, my mistake, it is the asdf package manager that adds this file.

soofstad (approved):

PR #352, opened 2023-02-13
feature(validation): optionally validate against provided type

### LINKED ISSUE #351: feature(dmss): validation type (optional)

soofstad:
- Allow to pass "as_type" to validation endpoint and validate the entity against this type (instead of type in entity)
- Using the "as_type" argument, also implicitly enables "allow_extra". 
  - Causing the validation to allow for extra attributes. 
  - Skip "valid child type" check
 
Be able to validate an entity against a "minimum valid blueprint" (required by UI Plugin)

closes #351

### DIFF
--- a/src/features/entity/use_cases/validate_entity.py
+++ b/src/features/entity/use_cases/validate_entity.py
@@ -44,12 +44,21 @@ def _recursive_validate_single_attribute(
 
 
 def validate_entity(
-    entity: dict | list, blueprint: Blueprint, get_blueprint: Callable[..., Blueprint], key: str = "^"
+    entity: dict | list,
+    blueprint: Blueprint,
+    get_blueprint: Callable[..., Blueprint],
+    key: str = "^",
+    allow_extra: bool = False,
 ) -> None:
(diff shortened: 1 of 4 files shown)


src/features/entity/use_cases/validate_entity.py
```diff
@@ -95,8 +108,10 @@ def validate_entity(
         )
 
 
-def validate_entity_use_case(entity: BasicEntity, user: User) -> str:
+def validate_entity_use_case(
+    entity: BasicEntity, user: User, as_type: common_type_constrained_string | None = None
```
ingeridhellen, line 112: No need for a default value, since that is already handled by api endpoint
```suggestion
    entity: BasicEntity, user: User, as_type: common_type_constrained_string | None
```

ingeridhellen (approved):
Looks good!

src/features/entity/use_cases/validate_entity.py
soofstad: fixed

PR #354, opened 2023-02-28
refactor: move validation out of use_case

### LINKED ISSUE #349: feat: add validation to CRUD endpoints

soofstad:

It adds validation when adding or updating entities to ensure that they have the correct attributes defined in their blueprint. The added validation revealed several buggy tests that had to be fixed.

Closes #349


src/common/utils/validators.py
```diff
@@ -35,125 +33,157 @@ def is_blueprint_instance_of(
...
-                raise ValidationException(f"Missing required attribute '{attribute.name}'", debug=debug_message)
+def _validate_primitive_attribute(attribute: BlueprintAttribute, value: bool | int | float | str, key: str):
+    python_type = BuiltinDataTypes(attribute.attribute_type).to_py_type()
+    if type(value) != python_type:
+        raise ValidationException(
+            f"Attribute '{attribute.name}' should be type '{python_type.__name__}'. Got '{type(value).__name__}'",
+            debug=_get_debug_message(key),
+        )
+    if not attribute.is_optional and python_type == str and value == "":
+        raise ValidationException(f"Missing required attribute '{attribute.name}'", debug=_get_debug_message(key))
```
soofstad, line 44: Don't think we should raise an exception here.
Empty string is valid for a required string attribute

soofstad (commented):
I can not request changes or approve my own PR.
However, look good! :D 
One thing, should we add the validation to the "add_raw/simple" endpoint?
I think we should as this the the endpoint often used during import.

src/common/utils/validators.py
```diff
@@ -35,125 +33,157 @@ def is_blueprint_instance_of(
...
+    if not entity.get("type"):
+        raise ValidationException("Every entity must have a 'type' attribute", debug=_get_debug_message(key))
+    _validate_entity(entity, get_blueprint, get_blueprint(entity["type"]), "exact", key)
 
 
 def validate_entity(
     entity: dict | list,
-    blueprint: Blueprint,
     get_blueprint: Callable[..., Blueprint],
+    blueprint: Blueprint,
```
KristianKjerstad, line 61: Could get_blueprint be used to get correct blueprint instead of this extra blueprint argument?

src/common/utils/validators.py
```diff
@@ -35,125 +33,157 @@ def is_blueprint_instance_of(
...
+    if not attribute.is_optional and python_type == str and value == "":
+        raise ValidationException(f"Missing required attribute '{attribute.name}'", debug=_get_debug_message(key))
 
-        return
-    if not attribute.is_optional and not value:
-        raise ValidationException(f"Missing required attribute '{attribute.name}'", debug=debug_message)
 
-    # Optional empty values are valid
-    if not value:
+def validate_entity_against_self(entity: dict | list, get_blueprint: Callable[..., Blueprint], key: str = "^") -> None:
```
KristianKjerstad, line 47: looks like validate_entity_against_self() and validate_entity() are very similar. Are there any guidelines when one should be used over the other?

ingeridhellen: I though it better with two functions, than one function with many optional input variables. Basically, use validate-entity_against_self if you want to check that the entity is correct according to its own type. If you want to check the entity towards the parent's attribute type or a minimum pattern, use validate_entity. You could use the latter function to solve the former issue if you wanted, but that would make the function call far more verbose.

src/common/utils/validators.py
ingeridhellen: Then you would use validate_entity_against_self.

src/common/utils/validators.py
KristianKjerstad: alright, cool!

KristianKjerstad (approved):

src/common/utils/validators.py
ingeridhellen: Had to update the tests too, since we had a test which was to fail on empty strings

soofstad:
:heavy_check_mark:

src/home/system/SIMOS/Link.json
```diff
@@ -3,6 +3,11 @@
   "description": "Reference to another entity.",
   "name": "Link",
   "attributes": [
+    {
+      "attributeType": "string",
+      "type": "dmss://system/SIMOS/BlueprintAttribute",
+      "name": "type"
+    },
```
netr0m, line 10:
```suggestion
    {
      "attributeType": "string",
      "type": "dmss://system/SIMOS/BlueprintAttribute",
      "name": "type"
    },
```
Is this necessary? Seeing as an identical attribute is defined in the [`system/SIMOS/Blueprint`](https://github.com/equinor/data-modelling-storage-service/blob/feat/validate-on-endpoints/src/home/system/SIMOS/Blueprint.json#L29,L33) blueprint, from which this blueprint inherits?

PR #363, opened 2023-03-07
refactor: move entity validator code into util

ingeridhellen:

Moves _recursive_validate_single_attribute and validate_entity into util/validators.py file. The functionality inside the functions is not changed

They will be used outside of the usecase, so it no longer makes sense to store them there

Refs #349


soofstad (approved):
I approve of this code. It seems very well written... :1st_place_medal:

PR #368, opened 2023-03-08
fix(add-to-path): allow adding packages to datasource

soofstad:
- If no "path" parameter in "DocumentService.add()", valid root packages will still be added to the data source

- Add support for creating root packages with the "add-to-path" endpoint

none, needed for https://github.com/equinor/dm-cli/issues/64


src/services/document_service.py
```diff
@@ -424,7 +425,21 @@ def _merge_entity_and_files(node: Node, files: Dict[str, UploadFile]):
                     )
 
     # Add entity by path
-    def add(self, data_source_id: str, path: str, document: Entity, files: dict, update_uncontained=False):
+    def add(self, data_source_id: str, path: str | None, document: Entity, files: dict, update_uncontained=False):
```
KristianKjerstad, line 428: would be nice to add some docstrings to functions in document_service.py, since its not so easy to understand what is happening from the function name alone.

Perhaps this should be a new issue.

KristianKjerstad (approved):

src/services/document_service.py
soofstad: Fixed!

PR #376, opened 2023-03-14
refactor: merge EntityType and Entity

ingeridhellen:

Replace EntityType with Entity, where the latter allow for extra attributes

Stig had already done this on the feat/validate-on-endpoints branch, so I thought I would separate it into its own PR.


soofstad (approved):
Nice!

PR #377, opened 2023-03-14
refactor: override entity assignment in Node class

ingeridhellen:

Fixes lint errors in Node class

Better to use types correctly than to ignore the warnings

### DIFF
--- a/src/domain_classes/tree_node.py
+++ b/src/domain_classes/tree_node.py
@@ -266,6 +266,7 @@ class Node(NodeBase):
         super().__init__(
             key, attribute, uid, parent, blueprint_provider, entity=entity, recipe_provider=recipe_provider
         )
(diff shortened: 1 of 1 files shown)


soofstad (approved):
If specialized classes is setting entity themself, should it not be removed from the parent class?
If that is possible ofc.
Can "NodeBase" do anything with the entity if it does not know it's type?

ingeridhellen:
> If specialized classes is setting entity themself, should it not be removed from the parent class? If that is possible ofc. Can "NodeBase" do anything with the entity if it does not know it's type?

NodeBase has an is_empty function which returns `not self.entity`, so it does use the variable. It also has a __repr__ function which assumes entity is a dict, that one should probably be moved into Node instead

PR #378, opened 2023-03-15
refactor: update document uses id_reference in path

### LINKED ISSUE #373: bug(get_document): attribute parameter in get document api call is ignored.
```
    def _get_by_id(self, document_id: str, depth: int = 1, attribute: str = ""):
        #TODO: Attribute seems to be redundent here
        params = {"depth": depth, "attribute": attribute}
        req = requests.get(
            f"{os.environ.get('DMSS_API', 'http://dmss:5000')}/api/documents/{document_id}", params=params,
            headers=self.headers  # type: ignore
        )  # type: ignore
        req.raise_for_status()

        return req.json()
```

It seems the intention is that the attribute of the document to be sent back. But that input is ignored.

soofstad:
- Refactor the "/document/" PUT (update) endpoint to be consistent with GET. That is moving "attribute" from a form param to be a part of the reference path parameter.

- Also; small change to make "NotFound" exceptions print stack in debug mode

- Endpoints should have one consistent way to reference documents

closes #373

### DIFF
--- a/src/features/document/document_feature.py
+++ b/src/features/document/document_feature.py
@@ -57,13 +57,11 @@ def get_by_path(
     return get_document_by_path_use_case(user=user, absolute_path=absolute_path)
 
 
-@router.put("/{data_source_id}/{document_id}", operation_id="document_update", responses=responses)
+@router.put("/{id_reference:path}", operation_id="document_update", responses=responses)
 @create_response(JSONResponse)
 def update(
-    data_source_id: str,
-    document_id: str,
+    id_reference: str,
     data: Json = Form(...),
-    attribute: Optional[str] = Form(None),
     files: Optional[List[UploadFile]] = File(None),
     update_uncontained: Optional[bool] = False,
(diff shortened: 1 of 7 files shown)


src/features/document/document_feature.py
```diff
@@ -57,24 +57,20 @@ def get_by_path(
...
-    data_source_id: str,
-    document_id: str,
+    id_reference: str,
     data: Json = Form(...),
-    attribute: Optional[str] = Form(None),
     files: Optional[List[UploadFile]] = File(None),
     update_uncontained: Optional[bool] = False,
     user: User = Depends(auth_w_jwt_or_pat),
 ):
     """Update document"""
```
KristianKjerstad, line 69:
```suggestion
    """Update document
        - **id_reference**: <data_source>/<document_uuid> (can also include an optional .<attribute> after <document_uuid>)
     """
```

to add more documentation to swagger ui

KristianKjerstad (approved):

PR #379, opened 2023-03-15
test: fix invalid entities in tests

ingeridhellen:

The work on #349 revealed several buggy tests. This PR solves all those I found

Refs #349


KristianKjerstad (approved):

src/tests/bdd/document/reference.feature
```diff
@@ -162,7 +161,8 @@ Feature: Add and remove references
     {
       "name": "myMooring",
       "type": "dmss://test-DS/TestData/Mooring",
-      "_id": "3f9ff99f-9cb5-4afc-947b-a3224eee341f"
+      "_id": "3f9ff99f-9cb5-4afc-947b-a3224eee341f",
+      "Bigness": 1
```
soofstad, line 165: This is wrong. If the validator forces us to send complete entites, when only inserting references, that must be fixed somehow.

src/tests/bdd/document/reference.feature
```diff
@@ -175,7 +175,8 @@ Feature: Add and remove references
       "Mooring": {
         "name": "myMooring",
         "type": "dmss://test-DS/TestData/Mooring",
-        "_id": "3f9ff99f-9cb5-4afc-947b-a3224eee341f"
+        "_id": "3f9ff99f-9cb5-4afc-947b-a3224eee341f",
+        "Bigness": 1
```
soofstad, line 179: Same

src/tests/bdd/document/update_with_blob.feature
```diff
@@ -88,6 +88,7 @@ Feature: Update document that has blob data
           "blob": {
             "name": "file1",
             "type": "dmss://system/SIMOS/Blob",
+            "_blob_id": "1",
```
soofstad, line 91: Think we should change the blueprint for "Blob", or create a special rule. "_blob_id" should be created on backend I believe

src/tests/unit/test_save.py
```diff
@@ -233,7 +233,6 @@ def test_save_no_overwrite_uncontained_document(self):
                         "_id": "2",
                         "name": "im_a_uncontained_attribute",
                         "type": "basic_blueprint",
-                        "contained": False,
```
soofstad, line 236: Don't think we can remove this. It's used at least when we are deleting stuff. To know if we should delete the "model contained" uncontained child.

soofstad (changes requested):

src/tests/unit/test_save.py
ingeridhellen: If an attribute is contained or not is defined in the blueprint, not in the entity, right? The `uncontained_blueprint` defines `uncontained_in_every_way`to be `contained: False`. Adding it here too makes no sense, right?

soofstad (approved):

PR #383, opened 2023-03-20
refactor: rename form plugin

KristianKjerstad:


soofstad (approved):

src/common/utils/get_storage_recipe.py
```diff
@@ -9,11 +9,18 @@
...
         "category": "view",
     }
 )
-default_form_edit = Recipe(**{"name": "Edit", "type": SIMOS.UI_RECIPE.value, "plugin": "form", "category": "edit"})
+default_form_edit = Recipe(
+    **{
+        "name": "Edit",
+        "type": SIMOS.UI_RECIPE.value,
+        "plugin": "@development-framework/dm-core-plugins/form",
+        "category": "edit",
```
eoaksnes, line 21: Shoud we remove the category?

KristianKjerstad: I cannot see that category is used anywhere in the dm-core-packages code.
should it be removed from UiRecipe blueprint in DMSS?

Issue #381, opened 2023-03-15
feature(API): clean up API interfaces to refere entities in a unified way 


l3abak:
There are many different ways to use refences. The interfaces should be clean up and two main referencing method should be kept. 
- by ID
a dotted path to the object consisting of 
`DataSource/{id}.child.children[0][2].child
`
This should directly work for children which are "contained" in "modelling" sense, regardless of containments status in "Storage" sense.  

The behaviour for items which are "no contained" in "model" needs discussion. The suggestion is to require explicit query to get the items which are non-contained on modelling level. 
`const id = get("DataSource/{id}.child.children[0][2].refchild.childID")
`
`get("{id}.child.child.value")
`
- by Path

Same as above, except instead of "id" the location of the entity in the path system is used. 
`DataSource/package1/package2/mycase.child.children[0][2].child
`

PR #384, opened 2023-03-20
refactor: standardize document 'by-path' endpoints parameters

soofstad:
- Standardize how parameters to "/document" and "/document-by-path/" endpoints should look like

- Easier to maintain and use endpoints if they all behave in a more similar fashion

related to #381

### DIFF
--- a/src/features/document/use_cases/remove_by_path_use_case.py
+++ b/src/features/document/use_cases/remove_by_path_use_case.py
@@ -1,14 +1,12 @@
-from typing import Optional
-
 from authentication.models import User
+from common.utils.string_helpers import split_dmss_ref
 from services.document_service import DocumentService
 from storage.internal.data_source_repository import get_data_source
 
 
-def remove_by_path_use_case(
-    user: User, directory: str, data_source_id: Optional[str] = None, repository_provider=get_data_source
-):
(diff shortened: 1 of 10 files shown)


src/features/document/use_cases/remove_by_path_use_case.py
```diff
@@ -1,14 +1,12 @@
...
 from authentication.models import User
+from common.utils.string_helpers import split_dmss_ref
 from services.document_service import DocumentService
 from storage.internal.data_source_repository import get_data_source
 
 
-def remove_by_path_use_case(
-    user: User, directory: str, data_source_id: Optional[str] = None, repository_provider=get_data_source
-):
+def remove_by_path_use_case(user: User, absolute_path: str, repository_provider=get_data_source):
```
KristianKjerstad, line 7: Is there a difference between `id_reference`and `absolute_path`? If not, can use id_reference here for consistency.
(For example in the remove_use_case.py, the function takes in `id_reference` and it looks like that is the same.)

KristianKjerstad (approved):
Looks good!
Just a question about function parameter for use case

eoaksnes (approved):
Nice. Was one comment fra Kristian.

PR #385, opened 2023-03-23
Fix/zip export and modelling errors

KristianKjerstad:
fix modelling errors and bug in zip export
causes dm-cli to fail when exporting: https://github.com/equinor/dm-cli/issues/69


src/storage/repositories/zip.py
```diff
@@ -24,7 +24,10 @@ def update(self, entity: dict, storage_recipe=None, **kwargs):
         entity.pop("_id", None)
         entity.pop("uid", None)
         entity["__path__"] = entity["__path__"].rstrip("/")
-        write_to = f"{entity['__path__']}/{entity['name']}.json"
+        if "name" not in entity:
+            write_to = f"{entity['__path__']}/unnamed_document.json"
```
soofstad, line 28: should this add some random id to allow for multiple unnamed_documents?

KristianKjerstad: ye!

PR #397, opened 2023-04-04
fix: bugs with new reference scheme

KristianKjerstad:
some bugs have been introduced in the new reference scheme.
This PR fixes some errors and update tests.
related to https://github.com/equinor/dm-cli/issues/81

### DIFF
--- a/src/common/tree_node_serializer.py
+++ b/src/common/tree_node_serializer.py
@@ -178,8 +178,13 @@ def tree_node_from_dict(
                     content_attribute.attribute_type = child["type"]
                     list_child_attribute = content_attribute
 
+                if "ref" in child:
+                    child_uid = child["ref"]
+                else:
+                    child_uid = child.get("_id", "")
+
                 list_child_node = tree_node_from_dict(
-                    uid=child.get("_id", ""),
+                    uid=child_uid,
                     entity=child,
                     key=str(i),
                     blueprint_provider=blueprint_provider,
--- a/src/common/tree_node_serializer.py
+++ b/src/common/tree_node_serializer.py
@@ -178,7 +178,7 @@ def tree_node_from_dict(
                     content_attribute.attribute_type = child["type"]
                     list_child_attribute = content_attribute
 
-                if "ref" in child:
+                if child["type"] == SIMOS.LINK.value:
                     child_uid = child["ref"]
                 else:
                     child_uid = child.get("_id", "")
--- a/src/common/tree_node_serializer.py
+++ b/src/common/tree_node_serializer.py
@@ -82,7 +82,7 @@ def tree_node_to_ref_dict(node: Node | ListNode) -> dict:
                 data[child.key] = [
                     {
                         "ref": child.uid,
-                        "targetType": child.type,
+                        "targetType": child.entity["targetType"] if child.type == SIMOS.LINK.value else child.type,
                         "targetName": child.name,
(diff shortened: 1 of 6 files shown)


src/common/tree_node_serializer.py
```diff
@@ -178,8 +178,13 @@ def tree_node_from_dict(
                     content_attribute.attribute_type = child["type"]
                     list_child_attribute = content_attribute
 
+                if "ref" in child:
+                    child_uid = child["ref"]
+                else:
+                    child_uid = child.get("_id", "")
+
```
soofstad, line 185: Should this rather do a check on type, and just assume "ref" is there if it's a "reference type"?

src/domain_classes/tree_node.py
```diff
@@ -266,12 +266,21 @@ def __init__(
...
+        ):
+            self.type = entity["targetType"]
 
     def is_root(self):
         return super().is_root()
 
     @property
     def name(self):
+        if "targetName" in self.entity and self.entity["type"] == "dmss://system/SIMOS/Link":
+            return self.entity.get("targetName", self.attribute.name)
```
soofstad, line 283: Is this really needed?
Where are we using the "name" property? In theory there should be no hard coded dependencies to it.
I vote for removing the whole property function

src/domain_classes/tree_node.py
```diff
@@ -266,12 +266,21 @@ def __init__(
         )
         self.entity: dict = entity if entity else {}
         self.error_message = None
+        if (
+            entity is not None
+            and entity != {}
+            and entity["type"] == "dmss://system/SIMOS/Link"
+            and "targetName" in entity
+        ):
+            self.type = entity["targetType"]
```
soofstad, line 275: Are we sure we want to "fake" which type this node is?
Seems like that could lead to some issues.
What kind of problems do we have that deems this necessary?

Also;
- Should use an enum for the "type" check,
- And "if entity" instead of "is not none" and "!= {}"

soofstad (approved):

src/common/tree_node_serializer.py
KristianKjerstad: sure!
that would make it a bit easier to read

src/domain_classes/tree_node.py
KristianKjerstad: I think removing it will require a bit of work.

the name property is used several places in document_service.

But I can add a new issue for it!

src/tests/unit/test_tree_node_helpers.py
```diff
@@ -980,69 +1063,19 @@ def test_to_dict(self):
...
-        engine_package_entity = {
-            "_id": "26d94353-3ff0-4b7d-bf06-86f006dd6f7b",
-            "type": "dmss://system/SIMOS/Package",
-            "name": "EnginePackage",
-            "isRoot": False,
-            "content": [engine_entity],
+        engine_package_node = get_engine_package_node()
+        engine_ref_node = engine_package_node.children[0].children[0]
+
+        assert engine_ref_node.storage_contained is True
```
KristianKjerstad, line 1069: before, this test used to check storage_contained is True.

but I think that is correct anymore after we added the new reference system (a package's content list contains references)

soofstad: Agreed, the test is wrong. Content of a package should not be storage contained, only model contained.

soofstad (approved):

PR #410, opened 2023-04-26
Refactor/more precise error for duplicate name

soofstad:
- Replace flak8 and pycln with "ruff" in pre-commit
-  Log stacktrace on every exception in debug mode
- Return correct target name on "duplicate name" error

- More useful error messages
- Standardized formatting and pre-commit across projects 

none


netr0m (changes requested):
I'd like for this PR to be split into several, focused pull requests.

Firstly, the pull request includes both changes to logic (e.g. stacktrace debug logging), as well as a new configuration for the pre-commit config. These changes are (or should be) unrelated.

Further, seeing as the new pre-commit config uses `ruff`, this leads to new rules being applied, which has caused a large amount of files to change. Sure, each change/file changed may be insignificant in terms of scale, but that's not really the point.

The various linting rules that have been applied as a result of this change, increases the difficulty of reviewing the pull request, as it reduces the visibility of the changes that are actually relevant to the topic/scope of the pull request (here, given by the title).

I suppose that the reviewers could filter by the commit relating the a given change (as long as the commits are scoped too, which they are in part here), but I would like it if we could adopt this as our flow regardless - one type of change => one branch => one PR. This has multiple benefits outside of simplifying the review process.

soofstad:
Added #noqa in responses.py to avoid migrating to ruff.

> I'd like for this PR to be split into several, focused pull requests.
> 
> Firstly, the pull request includes both changes to logic (e.g. stacktrace debug logging), as well as a new configuration for the pre-commit config. These changes are (or should be) unrelated.
> 
> Further, seeing as the new pre-commit config uses `ruff`, this leads to new rules being applied, which has caused a large amount of files to change. Sure, each change/file changed may be insignificant in terms of scale, but that's not really the point.
> 
> The various linting rules that have been applied as a result of this change, increases the difficulty of reviewing the pull request, as it reduces the visibility of the changes that are actually relevant to the topic/scope of the pull request (here, given by the title).
> 
> I suppose that the reviewers could filter by the commit relating the a given change (as long as the commits are scoped too, which they are in part here), but I would like it if we could adopt this as our flow regardless - one type of change => one branch => one PR. This has multiple benefits outside of simplifying the review process.

Added #noqa in responses.py to avoid migrating to ruff.

PR #411, opened 2023-04-27
Fix/wrong blueprint used to validate list items

soofstad:
- Fixes a "bug"/unintended behavior where integers were not valid values for "number" types
- Fixes a bug where a ListNode.blueprint returned the parent's blueprint

- Even "numbers" (floats in python) should be allowed to be whole numbers
- Validating with the wrong blueprint obviously gives ValidationException

none


netr0m (approved):
LGTM!

PR #413, opened 2023-05-02
refactor: get document and resolve references

eoaksnes:

* Use $ to mark references pointing to a document using id 
* Only have a single method to get a document (not several and utilities)


src/common/utils/resolve_reference.py
```diff
@@ -121,33 +127,49 @@ class AttributeItem:
...
+            path.append(path_element)
+    return resolve_reference_items(data_source, reference_items[1:], get_data_source, resolved_document, path)
+
+
+@dataclass(frozen=True)
+class ResolvedReference:
+    data_source_id: str
+    document_id: str
+    attribute_path: str
+    document: dict
```
soofstad, line 166:
```suggestion
    entity: dict
```

src/common/utils/resolve_reference.py
```diff
@@ -156,11 +178,19 @@ def resolve_reference(reference: str, data_source: DataSource, get_data_source:
...
+        reference = f"/{reference}"
 
     reference_items = reference_to_reference_items(reference)
-    document = resolve_reference_items(data_source, reference_items, get_data_source)
+    document, path = resolve_reference_items(data_source, reference_items, get_data_source)
     if not document:
         raise NotFoundException(
-            f"No document found that matches '{reference}', in the data source '{data_source.name}' could be found."
+            f"No document found that matches '{reference}', in the data source '{data_source.name}' could be found using {str(reference_items)}."
         )
```
soofstad, line 192: Does not resolve_reference_items() raise an error here? I think it should

soofstad (approved):
:crossed_fingers:

src/common/utils/resolve_reference.py
```diff
@@ -46,20 +46,19 @@ def reference_to_reference_items(reference: str, items=None, prev_deliminator=No
...
     elif prev_deliminator == "(" and deliminator == ")":  # By query
         items.append(QueryItem(query=content))
     elif prev_deliminator == "[" and deliminator == "]":  # By list index
         items.append(AttributeItem(f"[{content}]"))
     elif prev_deliminator == ".":  # By attribute
         items.append(AttributeItem(content))
-    else:  # Fallback to support the old reference format
-        items.append(IdItem(content))
+    else:
+        raise Exception("Not supported reference format")
```
KristianKjerstad, line 61: use ValidationException instead?
I believe that will give better error message for dm-cli / frontend.

Can also include  the reference as part of error message:
```python
raise Exception(f"Not supported reference format: {reference}")
```

src/services/document_service.py
```diff
@@ -183,39 +183,48 @@ def save(
...
+        document: dict = data_source.get(resolved_reference.document_id)
+
+        resolved_document: dict = resolve_document(
+            document,
+            data_source,
+            self.get_data_source,
+            resolved_reference.document_id,
+            depth + len(resolved_reference.attribute_path.split(".")),
+        )
+
```
KristianKjerstad, line 207: can add check on depth here and return resolved document if depth is 0 (or 1?? not sure what "indexing scheme" we use for depth - 0 or 1 indexing)

src/features/document/use_cases/get_document_by_path_use_case.py
```diff
@@ -14,20 +13,7 @@ def get_document_by_path_use_case(
...
-            raise NotImplementedError
-        case _:
-            raise NotImplementedError
-
-    if attribute:
-        document = document.get_by_path(attribute.split("."))
-
-    return tree_node_to_dict(document)
+    if "://" not in absolute_path:
+        absolute_path = f"/{absolute_path}"
```
KristianKjerstad, line 17: don't really understand why we have started to ignore the protocol.

KristianKjerstad (commented):
is the reference documentation still correct?
https://equinor.github.io/dm-docs/docs/concepts/blueprints#references

if not, we should update it soon

src/common/utils/resolve_reference.py
eoaksnes: Not always, created a separate task in the issue to better handle exceptions

src/services/document_service.py
eoaksnes: created a sperate task in the issue for this

PR #420, opened 2023-05-05
fix: let type of default in BlueprintAttribute be any

### LINKED ISSUE #365: feat: default attribute should be of type Any
BlueprintAttribute.json defines the attribute "default" to have `"attributeType": "string"`. This means that if we have an attribute with `"attributeType": "boolean"`, the tests will fail if we set `"default": false`

KristianKjerstad:
Introduce a new "any" type, that only should be used for the default attribute in BlueprintAttribute.json.
This means we dont have to write the default value as a string when modeling. Example, we can write

```json
{
      "attributeType": "boolean",
      "type": "dmss://system/SIMOS/BlueprintAttribute",
      "default": false
}
```

instead of
```json
{
      "attributeType": "boolean",
      "type": "dmss://system/SIMOS/BlueprintAttribute",
      "default": "false"
}
```
closes #365


src/common/utils/create_entity.py
```diff
@@ -38,14 +38,14 @@ def parse_value(attr: BlueprintAttribute, blueprint_provider):
...
-        if default_value is not None and len(str(default_value)) > 0 and attr.is_array:
+        if default_value is not None and len(str(default_value)) > 0 and attr.is_array and type not in PRIMITIVES:
             try:
                 return json.loads(default_value)
             except JSONDecodeError:
                 print(f"invalid default value: {default_value} for attribute: {attr}")
                 return []
 
-        if default_value == "":
+        if default_value is None:
```
KristianKjerstad, line 48: this change was necessary because the type of "default" in the BlueprintAttribute python class changed from string to `StrictStr | StrictInt | StrictFloat | StrictBool | list | dict | None`

src/common/tree_node_serializer.py
```diff
@@ -154,6 +154,8 @@ def tree_node_from_dict(
     for child_attribute in node.blueprint.get_none_primitive_types():
         if child_attribute.name == "_meta_":
             continue
+        if child_attribute.name == "default" and child_attribute.attribute_type == "any":
```
soofstad, line 157: Why do we not want to add attributes name default to the node?

src/common/utils/create_entity.py
```diff
@@ -69,21 +64,8 @@ def entity(self):
...
-
-    @staticmethod
-    def parse_json(attr: BlueprintAttribute):
-        try:
-            return json.loads(attr.default)
-        except JSONDecodeError:
-            print(f"invalid default value: {attr.default} for attribute: {attr}")
-            return ""
+        """A blueprint attribute is json if the default value is either a list or a dict."""
+        return type(attr.default) == dict or type(attr.default) == list
```
soofstad, line 68: Could be a bit more readable;
return attr.attributeType not in PRIMITIVE or attr.dimension != ""

src/common/utils/validators.py
```diff
@@ -143,6 +143,8 @@ def _validate_entity(
             _validate_primitive_attribute(
                 attributeDefinition, entity[attributeDefinition.name], f"{key}.{attributeDefinition.name}"
             )
+        elif attributeDefinition.attribute_type == "any" and attributeDefinition.name == "default":
```
soofstad, line 146:
```suggestion
        elif attributeDefinition.attribute_type == "any"
```
Do we only allow for "any" if the attribute is named "default"?

soofstad (changes requested):
Nice! Cool to see actual values in "default".
A few comments :slightly_smiling_face:

src/common/tree_node_serializer.py
KristianKjerstad: As far as I can see, this for loop (starting on line 154) is responsible for adding child nodes to the node.

Since the "default" attribute is not part of the PRIMITIVES enum 
(`PRIMITIVES = {"string", "number", "integer", "boolean"}`), the for loop will try to add "default" as a child node.

So in the same way that _meta_ attribute is ignored (line 155-156), we have to ignore the "default" attribute because we don't need to create a child node for it :)

src/common/utils/validators.py
KristianKjerstad: My thinking here was that we only want to allow "any" to a type for the "default" attribute so that users cannot use the "any" type when creating models.

Without going too much into the code, I would assume the user can manage skip the validation if they define blueprints with attributeType any. For example, I don't think the aValue attribute in the following blueprint would be validated if attributeType is set to "any"

```json
{
  "type": "dmss://system/SIMOS/Blueprint",
  "extends": ["dmss://system/SIMOS/NamedEntity"],
  "name": "ExampleUserBlueprint",
  "attributes": [
    {
      "name": "aValue",
      "attributeType": "any",
      "type": "dmss://system/SIMOS/BlueprintAttribute",
      "default": [1,2,3]
    }
  ]
}

```

KristianKjerstad:
@eoaksnes @ingeridhellen 
Stig did not get a change to finish review - would be great if one of you could take a look!

eoaksnes (approved):
Nice nice :)

PR #422, opened 2023-05-09
refactor: dont resolve references by default

soofstad:
Get_document
- Don't resolve links by default
- Default depth is "1"

- Posting/updating with data instead of references for links should not be possible. So this change will make that behavior more natural.

none


src/services/document_service.py
```diff
@@ -191,7 +191,8 @@ def save(
             return result
         return ref_dict
 
-    def get_document(self, reference: str, depth: int = 999) -> Node:
+    # TODO: Dont return Node. Doing this is 33% slower
```
eoaksnes, line 194: This should be done in the use case layer.

eoaksnes (approved):

PR #424, opened 2023-05-11
refactor: rewrite pretty_eq

ingeridhellen:

Rewrite pretty_eq and adapt all functions using it accordingly

The name "pretty_eq" is misleading, as pretty can mean both beautiful and fairly. Also, the function passes its input to traverse_compare in the wrong order. Finally, the try/catch is no longer needed, since traverse_compare no longer should raise Exceptions


soofstad (approved):
:+1:

PR #425, opened 2023-05-11
test: fix holes in traverse_compare

ingeridhellen:

Rewrite traverse_compare to something more readable and add checks for missing keys and indices
Add more tests
Fix tests now failing due to the change

traverse_compare didn't catch extra indices or keys

### DIFF
--- a/src/tests/unit/test_validate_entity.py
+++ b/src/tests/unit/test_validate_entity.py
@@ -221,11 +221,9 @@ class ValidateEntityTestCase(unittest.TestCase):
             "name": "myCarRental",
             "cars": [
                 {
-                    "type": "test_data/complex/CarTest",
+                    "type": "test_data/complex/RentalCar",
                     "name": "Volvo 240",
                     "plateNumber": "123",
-                    "wheels": [],
-                    "wheel": {"name": "Wheel", "power": 0.0, "type": "test_data/complex/WheelTest"},
                     "engine": {
                         "name": "myEngine",
                         "description": "Some description",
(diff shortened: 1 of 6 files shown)


src/tests/unit/test_validate_entity.py
```diff
@@ -221,11 +221,9 @@ def test_array_with_complex_child_getting_array_type_from_parent_node(self):
             "name": "myCarRental",
             "cars": [
                 {
-                    "type": "test_data/complex/CarTest",
+                    "type": "test_data/complex/RentalCar",
```
soofstad, line 224: Why change the type for this entity? That makes this into a "totally" different test.
Could rather add "RentalCar" as a new test. A few things in "CarTest" is not covered in "RentalCar"

soofstad (changes requested):

src/tests/unit/test_validate_entity.py
ingeridhellen: Yes, I know, that's the point. CarTest is used too many places and has too many non-optional attributes to be practical in here. Before, the CarRental.cars[x] entities in dmss would have all those non-optional attributes set to a default. Since pretty_eq ignored missing attributes, it didn't cause an error when comparing it to a much simpler CarTest object lacking those attributes. Now that I updated pretty_eq, it crashed as it should have long ago. Adding all the default params to the car seemed impractical, especially since the purpose of RentalCar testing was references.

soofstad: Right. Didn't see the "CarTest" was being used other places :+1:

soofstad (approved):

PR #428, opened 2023-05-15
feat: update the create lookup table use case to handle several to "recipe links" pacakges

KristianKjerstad:
update the create lookup table use case to handle several to "reciple links" pacakges.
https://github.com/equinor/dm-cli/issues/90

### DIFF
--- a/src/features/lookup_table/lookup_table_feature.py
+++ b/src/features/lookup_table/lookup_table_feature.py
@@ -21,18 +21,16 @@ router = APIRouter(tags=["default", "lookup-table"])
     responses={**responses},
 )
 @create_response()
-def create_lookup(
-    application: str, recipe_package_paths: list[str] = Query(), user: User = Depends(auth_w_jwt_or_pat)
-):
+def create_lookup(application: str, recipe_package: list[str] = Query(), user: User = Depends(auth_w_jwt_or_pat)):
     """
     Create a recipe lookup table from a package containing RecipeLinks.
     Associate it with an application.
     This can be used for setting Ui- and StorageRecipes for specific applications.
 
     - **application**: name of application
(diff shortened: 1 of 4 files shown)


ingeridhellen (approved):
Looks good! Could perhaps be nice with a bdd scenario that tests the effect of updating the lookup table though, ie first `/api/application/test-DS?recipe_package_paths=test-DS/root_package/recipe_links`, then `/api/application/test-DS?recipe_package_paths=test-DS/root_package/more_recipe_links`

src/features/lookup_table/lookup_table_feature.py
```diff
@@ -21,15 +21,18 @@
     responses={**responses},
 )
 @create_response()
-def create_lookup(recipe_package: str, application: str, user: User = Depends(auth_w_jwt_or_pat)):
+def create_lookup(
+    application: str, recipe_package_paths: list[str] = Query(), user: User = Depends(auth_w_jwt_or_pat)
```
soofstad, line 25: The URL looks weird to have "recipe_package_paths=one_path,recipe_package_paths=another_path".
Think we should either have this as one comma separated string, or multiple parameters "recipe_package". 
More important that the parameters and sensibly named on the outside, than in the controller.

soofstad (changes requested):

src/features/lookup_table/lookup_table_feature.py
KristianKjerstad: It does not seem like comma separated list is supported in fastAPI without writing our own code to extract the list elements (https://github.com/tiangolo/fastapi/issues/50)

renaming the variable instead.

soofstad (approved):

PR #429, opened 2023-05-16
fix: update_documents now supports new reference scheme

soofstad:
- DocumentService.update() now understands "ReferenceItems" when getting the "parentNode" (needed since the main target might not exist
- New utility function to split a reference into the different components
- Moved "split_data_source_and _reference" to separate function, since it's needed outside of "resolve_reference()"
- Fixed a bug in unittests where "save()" where never actually tested, since we passed documents by reference. We now operate on a copy, so that "save()" is properly tested.

- Update failed to operate on items in a list (the bug in the tests caused this to go undiscovered)
- The rest follows from that

none

### DIFF
--- a/src/common/utils/resolve_reference.py
+++ b/src/common/utils/resolve_reference.py
@@ -10,6 +10,31 @@ from enums import Protocols
 from storage.data_source_class import DataSource
 
 
+def split_reference(reference: str) -> list[str]:
+    """Splits a reference into it's string components
+    e.g. "$123.content[1]" -> ["$123", ".content", "[1]"]"""
+    parts = []
+    remaining_ref = reference
+    while remaining_ref:
+        if remaining_ref[0] in "./":
+            prefix_delim = remaining_ref[0]
+            content = re.split(r"[./\[(]", remaining_ref, 2)[1]
+            parts.append(f"{prefix_delim}{content}")
+            remaining_ref = remaining_ref.removeprefix(parts[-1])
+            continue
+        if remaining_ref[0] in "[(":
+            prefix_delim = remaining_ref[0]
+            content = re.split(r"[\[\]\(\)]", remaining_ref, 2)[1]
+            parts.append(f"{prefix_delim}{content}{']' if prefix_delim == '[' else ')'}")
+            remaining_ref = remaining_ref.removeprefix(parts[-1])
+            continue
+
+        content = re.split(r"[./]", remaining_ref, 1)[0]
+        remaining_ref = remaining_ref.removeprefix(content)
+        parts.append(content)
+    return parts
+
+
 def _next_reference_part(reference: str) -> Tuple[str, Union[str, None], str]:
     """Utility to get next reference part."""
     content = reference  # Default to reference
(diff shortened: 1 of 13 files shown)


src/common/utils/resolve_reference.py
```diff
@@ -10,6 +10,31 @@
...
 
+def split_reference(reference: str) -> list[str]:
+    """Splits a reference into it's string components
+    e.g. "$123.content[1]" -> ["$123", ".content", "[1]"]"""
+    parts = []
+    remaining_ref = reference
+    while remaining_ref:
+        if remaining_ref[0] in "./":
+            prefix_delim = remaining_ref[0]
+            content = re.split(r"[./\[(]", remaining_ref, 2)[1]
```
ingeridhellen, line 21: What if `reference=="/[(_id=1)].content[(name=myCarRental)].cars[0]"`. Then `content==""`. Perhaps add some tests for this?

eoaksnes (approved):

src/common/utils/resolve_reference.py
```diff
@@ -26,50 +51,15 @@ def _next_reference_part(reference: str) -> Tuple[str, Union[str, None], str]:
...
-
 @dataclass
 class IdItem:
     """Points to an id."""
 
     id: str
 
-    def resolve(self, document: dict | list, data_source: DataSource, get_data_source: Callable) -> Tuple[dict, str]:
+    def resolve(
+        self, document: dict | list | None, data_source: DataSource, get_data_source: Callable
```
ingeridhellen, line 61: document isn't used anywhere in the function

src/common/utils/resolve_reference.py
```diff
@@ -137,7 +127,9 @@ class AttributeItem:
     def __repr__(self):
         return f'Attribute="{self.path}"'
 
-    def resolve(self, document: dict | list, data_source: DataSource, get_data_source: Callable) -> Tuple[dict, str]:
+    def resolve(
+        self, document: dict | list | None, data_source: DataSource, get_data_source: Callable
+    ) -> tuple[dict | list | None, str]:
         try:
             result = find(document, [self.path])
```
ingeridhellen, line 134: find does not accept document of type None

src/common/utils/resolve_reference.py
```diff
@@ -167,6 +214,8 @@ def resolve_reference_items(
         path = [path_element]
     else:
         if resolved_document and "_id" in resolved_document:
+            if isinstance(resolved_document, list):
```
ingeridhellen, line 217: How can the resolved document be both a list and have _id at the same time?

src/common/utils/resolve_reference.py
```diff
@@ -10,6 +10,31 @@
 from storage.data_source_class import DataSource
 
 
+def split_reference(reference: str) -> list[str]:
+    """Splits a reference into it's string components
+    e.g. "$123.content[1]" -> ["$123", ".content", "[1]"]"""
```
ingeridhellen, line 15: Is it wise to introduce yet another split reference format, here with "."? Should we not agree on one format internally in dmt and stick to it?

src/services/document_service.py
```diff
@@ -116,12 +121,15 @@ def save(
         path="",
         update_uncontained: bool = False,
         combined_document_meta: dict | None = None,
+        initial: bool = False,  # Should never be passed to children
```
ingeridhellen, line 124: I don't understand this param. Perhaps add more info to the docstring?

src/common/utils/resolve_reference.py
soofstad: Not my change. I believe this was done so all the "ReferenceItems" could have the same interfaces

src/common/utils/resolve_reference.py
soofstad: The problem is that I need the "string parts". Not a list of RefereceItems.

I tried to reuse the "_next_reference_part()", but it was returning all kinds of stuff like empty string, single dots and so on. Also it removes the prefix, like ".", "[" and "(". 

But I agree, this should not be duplicated. Can see if we can make this work

src/common/utils/resolve_reference.py
soofstad: fixed in https://github.com/equinor/data-modelling-storage-service/pull/435

src/common/utils/resolve_reference.py
soofstad: fixed in https://github.com/equinor/data-modelling-storage-service/pull/435

src/services/document_service.py
soofstad: more description added in https://github.com/equinor/data-modelling-storage-service/pull/435

Issue #414, opened 2023-05-03
refactor: reduce/merge number of endpoints


eoaksnes:
- [x] Merge document get by id and get by path @eoaksnes 
- [x] Change delete endpoint @eoaksnes (@soofstad)
  - [x]  Is it possible to delete things with format [[1,2],[3,4]] @ingeridhellen 
- [x] Merg document add endpoints @eoaksnes (@KristianKjerstad)  https://github.com/equinor/data-modelling-storage-service/pull/430
- [x] use new reference format for endpoints in export_feature.py @KristianKjerstad  https://github.com/equinor/data-modelling-storage-service/pull/441

KristianKjerstad:
Question regarding Merge document add endpoint:


Example: we want to add a document to a root package with id 100. Assume the root package has 6 documents initially.


Example of valid requests are:
`POST localhost:5000/api/documents/data-source-name/$100`
`POST localhost:5000/api/documents/data-source-name/root_package`

However, should we also support  adding document using the `.attribute` syntax?
`POST localhost:5000/api/documents/data-source-name/root_package.content[6]`

soofstad:
Been a small discussion on this already.

IMO, the last request should be done as an "update()", that endpoint handles non-existing items/attributes.
Adding only makes sense in context of a list.

@ingeridhellen proposed we change "add()" to "insert(index=-1)" and I think that could be a good idea.

Question: Is there any scenario where you need to "add()" NOT to a list, and can't use "update()"?

KristianKjerstad:
okay, cool!

It makes sense to use "update" instead in the last request.

I can't think of any scenario where we need to add not to a list and can't use update

KristianKjerstad:
from discussion, it was concluded that the export endpoints should only handle reference on path format.

PR #430, opened 2023-05-16
refactor/merge document add endpoints

KristianKjerstad:
Merge document add endpoints into a single endpoint, that takes in a reference.

#414


src/features/document/use_cases/add_document_to_path_use_case.py
```diff
@@ -3,26 +3,29 @@
...
     repository_provider=get_data_source,
 ):
     document_service = DocumentService(repository_provider=repository_provider, user=user)
     request_document: Entity = Entity(**document)
-    data_source_id, path, attribute = split_dmss_ref(path_reference)
+    if "/" not in reference:
+        # Reference only contains a data source
+        pass
+    elif "://" not in reference:
+        reference = f"/{reference}"
```
KristianKjerstad, line 25: TODO - investigate if there is a better way to handle add document to data source?

src/tests/bdd/document/add_document.feature
```diff
@@ -222,29 +220,109 @@ Feature: Add document with document_service
...
+      {
+        "type": "dmss://data-source-name/root_package/Operation",
+        "name": "operation3",
+        "description": "",
+        "phases": []
+      }
+    }
+    """
+    Then the response status should be "OK"
+    Then the response status should be "OK"
```
KristianKjerstad, line 251: duplicate line can be removed

src/services/document_service.py
```diff
@@ -454,78 +362,91 @@ def _merge_entity_and_files(node: Node, files: Dict[str, BinaryIO]):
...
+            return self._add_document_to_data_source(reference, document, update_uncontained)
 
-        for node in target.traverse():
-            print(node.uid)
+        target: Node = self.get_document(reference, resolve_links=True, depth=99)
 
+        data_source_id, _reference = split_data_source_and_reference(reference)
         if not target:
-            raise NotFoundException(f"Could not find '{path}' in data source '{data_source_id}'")
+            raise NotFoundException(f"Could not find '{reference}' in data source '{data_source_id}'")
```
KristianKjerstad, line 411: A very small detail, but it might be nice to move the "if not target" part above line 409 - so that the check on "target" is right after getting av value for "target"

KristianKjerstad (commented):
👍

Looks good! But can't approve since i'm also author

src/services/document_service.py
eoaksnes: Need to get the data_source_id from the reference before raising NotFoundException. I don't think it is possible to move it up there.

ingeridhellen (approved):
A lot of good stuff here. Just a bit worried about how the references are parsed

src/services/document_service.py
```diff
@@ -454,78 +362,91 @@ def _merge_entity_and_files(node: Node, files: Dict[str, BinaryIO]):
...
-            except NotFoundException:
-                pass
-            document_repository = self.repository_provider(data_source_id, self.user)
-            document_repository.update(document_dict)
-            return {"uid": document_dict["_id"]}
+        validate_entity_against_self(document, self.get_blueprint)
+        entity: Entity = Entity(**document)
 
-        target: Node = self.get_document(f"/{data_source_id}/{path}", resolve_links=True, depth=99)
+        if "/" not in reference:  # We're adding something to the dataSource itself
```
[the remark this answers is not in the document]
eoaksnes: The split_data_source_and_reference crashes if the references only contain a reference to the data source.

But this can be a good idea, I will add it to the to-do list for later, but not include it in this PR to make it bigger.

src/services/document_service.py
KristianKjerstad: oh, right. Resolved!

Issue #389, opened 2023-03-23
implement the new referencing system with query and array.


l3abak:
## Description

_A clear and concise description of what the task is about._

```
As a user
I want to create references with queries and arrays
So that I can reference contained entities that does not have an storage generated id
```

This is part of the Epic https://github.com/equinor/data-modelling-storage-service/issues/374.

## Acceptance criteria

_List what the task must do/contain before the product owner/interests accept it as complete._

- [ ] todo
 
## Breakdown

_Include specific sub-tasks in the order they need to be done._

- [x] test: add BDD and/or unit test for get document using different depths @ingeridhellen 
- [x] test: add BDD and/or unit test for get document using wrong references @ingeridhellen 
- [x] test: add BDO and/or unit test for resolve links and get document @ingeridhellen 
- [x] refactor: use class for Reference instead of str @ingeridhellen 
- [x] refactor: see https://github.com/equinor/data-modelling-storage-service/pull/457#discussion_r1236723871

#### Absolute references

```
path: 

{protocol}://{data source name}/{name=root_package_name, type=Package, root:true}.content{name=package_name,  type=Package}.content{name=blueprint_name,  type=Blueprint}

dmss://system/SIMOS/Blueprint
dmss+qr://system/{name:SIMOS, type=Package, root:true}.content{name:Blueprint,  type=Blueprint}


dmss+qr://appDS/{name:instances, type=Package}.content{name:myrental}.cars[0].engine
dmss://appDS/instances/myrental.cars[0].engine
or
dmss+qr://appDS/{_id:-myrental-id-}.cars[0].engine
dmss+id://appDS/-myrental-id-.cars[0].engine

dmss://appDS/models/carParts/Wheel
dmss+qr://appDS/{name:models}.content{name:carParts}.content{name:Wheel}
or
dmss+qr://appDS/{_id:-Wheel-id-}
dmss+id://appDS/-Wheel-id-
```

#### Root Path

Addressing elements on the same root (protocol and data storage).

```
/{_id:-myrental-id-}.cars[0].engine
/{name:instances}.content{name:myrental}.cars[0].engine


/{_id:-myrental-id-}
```

#### Relative references

all elements contained in one document, i.e. one contained object.

```
inside CarRentalCompany.cars[0]
../../customers[0]
../../customers[0]{name:"John Smith"}

```

#### StorageReferences 

Could have the following forms:

```
{protocol}://{data source name}/id

[...]

PR #431, opened 2023-05-19
test: add get_document with depth tests

ingeridhellen:

Add testing of the depth parameter. Not everything worked well, so had to do some rewriting.

We need to test that the depth parameter works as intended

Refs #389


soofstad (approved):
Good!

PR #434, opened 2023-05-19
test: restructure get document tests

ingeridhellen:

moves depth testing into separate fail to allow only on assert in each test (the logic in test_inputs have not been altered)
deletes test_get_complete_nested_reference and test_get_complete_document, as I don't think they give us anything new of value

Makes it easier to make more tests on this

Refs #389


KristianKjerstad (commented):
good cleanup!

However, I don't have enough knowledge about this part of the code to tell if removing test_get_complete_document() and test_get_complete_document() is a good idea, so perhaps someone else also should give a review before merge.

soofstad (approved):

PR #437, opened 2023-05-23
refactor: remove redundant delete endpoint

soofstad:


KristianKjerstad:
https://github.com/equinor/data-modelling-storage-service/blob/master/src/tests/bdd/document/remove_by_path.feature#L60


https://github.com/equinor/data-modelling-storage-service/blob/master/src/tests/bdd/document/remove_by_path.feature#L102

PR #440, opened 2023-05-24
feat: attribute-selector as default initial recipe

### LINKED ISSUE #360: feat: set default initial recipe
The default should be an AttributeSelector plugin with default as self Yaml and the rest should be using Edit (form)

soofstad:
- Adds 'attribute-selector' as a default initialUiRecipe

DM framework should work with sane defaults

closes #360


ingeridhellen (approved):
Nice!

PR #441, opened 2023-05-30
fix: updated reference format used in export endpoints

KristianKjerstad:
#414


src/features/export/use_cases/export_meta_use_case.py
```diff
@@ -69,15 +70,34 @@ def _collect_entity_meta_by_path(
...
+    document_service = DocumentService(user=user)
+    if "://" not in reference:
+        reference = f"/{reference}"
+    resolved_reference: ResolvedReference = resolve_reference(reference, document_service.get_data_source)
+    data_source_id = resolved_reference.data_source_id
     data_source = get_data_source(data_source_id, user)
 
-    path_elements = path.split("/")
-    root_package_name = path_elements.pop(0)
+    path_elements = reference.strip("/").split("/")
```
ingeridhellen, line 91: What if reference has protocol? Perhaps use reference_to_reference_items instead?

ingeridhellen (changes requested):

PR #443, opened 2023-05-30
Refactor/merge remove endpoints

eoaksnes:

* Update/merge "delete_document" to handle new reference format
* resolve_reference() now accepts param "resolve", used to decide whether or not to resolve the target of a reference if it's a Reference type-

* Support new reference format

Rebase of https://github.com/equinor/data-modelling-storage-service/pull/437 (keep history)


src/common/utils/delete_documents.py
```diff
@@ -14,6 +14,25 @@ def delete_list_recursive(value: Union[list, dict], data_source: DataSource):
...
 
 
+def delete_by_attribute_path(document: dict, path: list[str], data_source: DataSource, get_data_source) -> dict:
+    target = document
+    path_elements = [e.strip("[]./") for e in path]
+    for element in path_elements[:-1]:  # Step through all the path items except the last one that should be deleted
+        if isinstance(target, list):
+            target = target[int(element)]
+        else:
+            target = target[element]
```
ingeridhellen, line 24: Couldn't you use find (src/common/utils/data_structure/find.py) for this?

src/common/utils/delete_documents.py
```diff
@@ -14,6 +14,25 @@ def delete_list_recursive(value: Union[list, dict], data_source: DataSource):
...
+
+    if isinstance(target, list):
+        delete_dict_recursive(target[int(path_elements[-1])], data_source=data_source)
+        del target[int(path_elements[-1])]
+    else:
+        delete_dict_recursive(target[path_elements[-1]], data_source=data_source)
```
ingeridhellen, line 30: How do you know that `target[path_elements[-1]` is a dict? The function has no docstring stating that the path must lead to a dict and not a list

src/common/utils/resolve_reference.py
```diff
@@ -103,16 +100,23 @@ def get_entry_point(self, data_source: DataSource) -> Tuple[dict, str]:
...
         return result[0], result[0]["_id"]
 
-    def get_child(self, document: dict | list, data_source: DataSource, get_data_source: Callable) -> Tuple[Any, str]:
+    def get_child(
+        self, document: dict | list, data_source: DataSource, get_data_source: Callable, resolve: bool = False
+    ) -> Tuple[Any, str]:
         # Search inside an existing document (need to resolve any references first before trying to compare against filter)
         elements = [
-            resolve_reference(f"/{data_source.name}/{f['address']}", get_data_source).entity if is_reference(f) else f
+            resolve_reference(f"/{data_source.name}/{f['address']}", get_data_source, resolve).entity
```
ingeridhellen, line 108: We should always resolve the references before searching in them, I think. Or can you think of a reason why we'd need queries like `[(type=reference)]`?

src/common/utils/resolve_reference.py
```diff
@@ -124,9 +128,11 @@ class AttributeItem:
     path: str
 
     def __repr__(self):
-        return f'Attribute="{self.path}"'
+        return self.path
 
-    def get_child(self, document: dict | list, data_source: DataSource, get_data_source: Callable) -> tuple[Any, str]:
+    def get_child(
+        self, document: dict | list, data_source: DataSource, get_data_source: Callable, resolve: bool = False
```
ingeridhellen, line 134: This function does not use the resolve input param. So the param only has effect on queries? That's unintuitive.
I don't like this new resolve parameter. I think resolve_reference never should return a resolved document. It should only resolve while nesting downwards. If the caller needs to get the entity resolved it can do so itself easily

src/services/document_service.py
```diff
@@ -318,32 +284,21 @@ def update_document(
...
-            child_node_ids = (child.node_id for child in parent.children[0].children if child.uid == sub.document_id)
-            child_node_id = next(child_node_ids)
-
-            # The first child of a directory is always 'content'
-            parent.children[0].remove_by_child_id(child_node_id)
-            self.save(parent, data_source_id)
-            delete_document(data_source, document_id=sub.document_id)
+        resolved_reference: ResolvedReference = resolve_reference(reference, self.get_data_source)
+        # If the reference goes through a parent, get the parent document
+        document = data_source.get(resolved_reference.document_id)
```
ingeridhellen, line 293: Could you move this inside the if sentence? There is no need to get the document if there is no attribute path.

KristianKjerstad:
I think 2 tests must be updated: 

https://github.com/equinor/data-modelling-storage-service/blob/master/src/tests/bdd/document/remove_by_path.feature#L60

https://github.com/equinor/data-modelling-storage-service/blob/master/src/tests/bdd/document/remove_by_path.feature#L102

ingeridhellen (changes requested):

eoaksnes:
> I think 2 tests must be updated:
> 
> https://github.com/equinor/data-modelling-storage-service/blob/master/src/tests/bdd/document/remove_by_path.feature#L60
> 
> https://github.com/equinor/data-modelling-storage-service/blob/master/src/tests/bdd/document/remove_by_path.feature#L102

They are updated, you look at the master branch :)

src/features/document/document_feature.py
```diff
@@ -11,12 +11,18 @@
...
 from .use_cases.update_document_use_case import update_document_use_case
 
 router = APIRouter(tags=["default", "document"], prefix="/documents")
 
+"""    - **reference**: A reference to a package or a data source
+      - By id: PROTOCOL://DATA SOURCE/$ID.Attribute
+      - By path: PROTOCOL://DATA SOURCE/ROOT PACKAGE/SUB PACKAGE/ENTITY.Attribute
+      - By query: PROTOCOL://DATA SOURCE/$ID.list(key=value)
+
+      The PROTOCOL is optional, and the default is dmss."""
```
KristianKjerstad, line 24: Looks like this comment about reference is moved out from function docstring. The advantage with keeping it in the docstring is that it will be included in the swagger documentation  - I would prefer to move it back into the function docstring

src/common/utils/resolve_reference.py
```diff
@@ -237,8 +246,10 @@ class ResolvedReference:
     entity: dict | list
 
 
-def resolve_reference(reference: str, get_data_source: Callable) -> ResolvedReference:
-    """Resolve the reference into a document."""
+def resolve_reference(reference: str, get_data_source: Callable, resolve: bool = False) -> ResolvedReference:
+    """Resolve the reference into a document.
+
+    resolve: Resolve the target if it's a reference"""
```
KristianKjerstad, line 252: The extra "resolve" argument is q bit confusing. 
I'm not sure what it means to call resolve_reference() with resolve = False? That document returned in the  ResolvedReference is not a dict?

src/common/utils/get_blueprint.py
```diff
@@ -18,6 +19,12 @@ def get_blueprint(self, type: str) -> Blueprint:
         resolved_reference: ResolvedReference = resolve_reference(
             type, lambda data_source_name: get_data_source(data_source_name, self.user)
         )
+        if is_reference(resolved_reference.entity):
+            resolved_reference = resolve_reference(
+                f"{resolved_reference.data_source_id}/{resolved_reference.entity['address']}",
```
ingeridhellen, line 24: sometimes the address already contains a datasource. ex L70 in test_reference_resolve

src/common/utils/get_blueprint.py
```diff
@@ -18,6 +19,12 @@ def get_blueprint(self, type: str) -> Blueprint:
...
+                lambda data_source_name: get_data_source(data_source_name, self.user),
+            )
+            return Blueprint(resolved_reference.entity, type)
```
ingeridhellen, line 27: you can remove this line

src/common/utils/get_resolved_document_by_id.py
```diff
@@ -73,6 +73,10 @@ def get_complete_sys_document(
         address = f"{data_source.name}/{address}"
 
     resolved_reference: ResolvedReference = resolve_reference(address, get_data_source)
+    if is_reference(resolved_reference.entity) and resolve_links:
+        resolved_reference = resolve_reference(
+            f"{resolved_reference.data_source_id}/{resolved_reference.entity['address']}", get_data_source
```
ingeridhellen, line 78: the address may contain datasource already

src/common/utils/resolve_reference.py
```diff
@@ -102,14 +102,17 @@ def get_entry_point(self, data_source: DataSource) -> Tuple[dict, str]:
         return result[0], result[0]["_id"]
 
     def get_child(self, document: dict | list, data_source: DataSource, get_data_source: Callable) -> Tuple[Any, str]:
+        if isinstance(document, dict) and is_reference(document):
+            document = resolve_reference(f"/{data_source.name}/{document['address']}", get_data_source).entity
```
ingeridhellen, line 106: same

ingeridhellen (approved):

src/common/utils/delete_documents.py
ingeridhellen: Added as subissue in https://github.com/orgs/equinor/projects/189/views/3?reload=true&pane=issue&itemId=27214325

PR #448, opened 2023-06-01
fix: delete nested list

ingeridhellen:

- refactors test_remove to use a shared setup
- add test for deleting a list attribute
- fix issues revealed by the above mentioned test

It should be possible to delete attributes of type list

Refs #414

### DIFF
--- a/src/common/utils/delete_documents.py
+++ b/src/common/utils/delete_documents.py
@@ -5,32 +5,39 @@ from enums import REFERENCE_TYPES, SIMOS
 from storage.data_source_class import DataSource
 
 
-def delete_list_recursive(value: Union[list, dict], data_source: DataSource):
+def _delete_list_recursive(value: Union[list, dict], data_source: DataSource):
     """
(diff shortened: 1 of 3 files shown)


src/common/utils/delete_documents.py
```diff
@@ -5,32 +5,39 @@
 from storage.data_source_class import DataSource
 
 
-def delete_list_recursive(value: Union[list, dict], data_source: DataSource):
+def _delete_list_recursive(value: Union[list, dict], data_source: DataSource):
```
KristianKjerstad, line 8: are the functions that starts with "_" not meant to be used outside this file?

KristianKjerstad (approved):
good!

src/common/utils/delete_documents.py
ingeridhellen: Yes. I think that is a python naming convention. Though now that I googled it, it seems to be used mainly for classes and it should actually be 2 underscores, not just one. Still, I've seen it used in the same way as I have now several places in dmss, so I think it is fine.

PR #456, opened 2023-06-12
fix: tree_node_from_dict() should not include optional attributes if …

KristianKjerstad:
…not in the entity


src/services/document_service.py
```diff
@@ -253,28 +253,76 @@ def get_document(self, reference: str, depth: int = 0, resolve_links: bool = Fal
...
+                    entity=node_entity,
+                )
+            else:
+                node = Node(
+                    key=attribute_to_update,
+                    entity=node_entity,
+                    attribute=attribute,
+                    blueprint_provider=self.get_blueprint,
+                )
+        node.parent = parent_node
```
ingeridhellen, line 297: `node` can still be none if `attribute_to_update not in parent_blueprint_attribute_names`. You should probably add some error handling here.

If I'm not wrong, there are two valid cases here.
1. `node is not None`
2. `node is None and attribute_to_update in [attribute.name for attribute in parent_node.blueprint.attributes if attribute.is_optional]`

All other cases should give an error.

src/services/document_service.py
```diff
@@ -253,28 +253,76 @@ def get_document(self, reference: str, depth: int = 0, resolve_links: bool = Fal
...
+        parent_node.children.append(node)
```
ingeridhellen, line 298: Why?

src/services/document_service.py
```diff
@@ -253,28 +253,76 @@ def get_document(self, reference: str, depth: int = 0, resolve_links: bool = Fal
...
+            and attribute_to_update in parent_blueprint_attribute_names
+        ):
+            attribute = [
+                attribute for attribute in parent_node.blueprint.attributes if attribute.name == attribute_to_update
+            ][0]
+
+            # We only want to add the attribute if it does not already exist AND is optional.
+            if not attribute.is_optional:
+                raise ValidationException(f"Could not update node. attribute '{attribute.name}' is not optional.")
+            if attribute.dimensions.dimensions != [""]:
```
ingeridhellen, line 283:
```suggestion
            if attribute.is_array():
```

ingeridhellen (approved):

PR #457, opened 2023-06-19
refactor: reference object

ingeridhellen:

- The Reference class was renamed to ReferenceEntity
- The new Reference class was created
- Started using it in resolve_reference and adjacent functions. Can probably be added more places.

The reference string could have multiple formats, and it was often not documented which format was needed.

Refs #389

### DIFF
--- a/src/common/address.py
+++ b/src/common/address.py
@@ -17,9 +17,9 @@ class Address:
 
     @classmethod
     def from_absolute(cls, address: str):
-        """Returns an instance of the Reference class based on the reference input
+        """Returns an instance of the Address class based on the address input
 
-        @param reference: Must be on one of the following formats (.Attribute is optional)
+        @param address: Must be on one of the following formats (.Attribute is optional)
             - dmss://DATA_SOURCE/(PATH|$ID).Attribute
             - DATA_SOURCE/(PATH|$ID).Attribute
             - /DATA_SOURCE/(PATH|$ID).Attribute
--- a/src/common/address.py
+++ b/src/common/address.py
@@ -19,10 +19,10 @@ class Address:
     def from_absolute(cls, address: str):
         """Returns an instance of the Reference class based on the reference input
 
-        @param reference: Must be on one of the following formats
-            - dmss://DATA_SOURCE/(PATH|ID).Attribute
-            - DATA_SOURCE/(PATH|ID).Attribute
-            - /DATA_SOURCE/(PATH|ID).Attribute
+        @param reference: Must be on one of the following formats (.Attribute is optional)
+            - dmss://DATA_SOURCE/(PATH|$ID).Attribute
+            - DATA_SOURCE/(PATH|$ID).Attribute
+            - /DATA_SOURCE/(PATH|$ID).Attribute
         """
         protocol = Protocols.DMSS.value
         path = None
--- a/src/common/address.py
+++ b/src/common/address.py
@@ -2,7 +2,7 @@ from common.exceptions import ApplicationException
 from enums import Protocols
 
 
-class Reference:
+class Address:
     def __repr__(self):
         path = f"/{self.path}" if self.path else ""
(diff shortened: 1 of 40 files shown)


src/common/reference.py
```diff
@@ -0,0 +1,48 @@
+from common.exceptions import ApplicationException
+from enums import Protocols
+
+
+class Reference:
```
ingeridhellen, line 5: Should I perhaps rename it to Address?

src/tests/bdd/document/reference.feature
```diff
@@ -54,7 +54,7 @@ Feature: Add and remove references
       "isRoot": true
     }
     """
-    Given i access the resource url "/api/reference/test-DS/$1.content.0"
+    Given i access the resource url "/api/reference/test-DS/$1.content[0]"
```
ingeridhellen, line 57: This might cause some issues in frontend, but it is more consistent with the other API endpoints

src/services/document_service.py
```diff
@@ -206,7 +206,7 @@ def save(
         return ref_dict
 
     # TODO: Dont return Node. Doing this is ~33% slower
-    def get_document(self, reference: str, depth: int = 0, resolve_links: bool = False) -> Node | ListNode:
+    def get_document(self, address: Address, depth: int = 0, resolve_links: bool = False) -> Node | ListNode:
```
KristianKjerstad, line 209: Could update function docstring: "reference" -> "address"

src/services/document_service.py
```diff
@@ -306,7 +305,7 @@ def _get_node_to_update(self, reference: str, node_entity: Union[dict, list]) ->
 
     def update_document(
         self,
-        reference: str,
+        address: Address,
```
KristianKjerstad, line 308: could update docstring:

"What to update is referred to with a reference string." -> "What to update is referred to with an address."

src/common/utils/resolve_reference.py
```diff
@@ -237,20 +236,19 @@ class ResolvedReference:
     entity: dict | list
 
 
-def resolve_reference(reference: str, get_data_source: Callable) -> ResolvedReference:
+def resolve_reference(address: Address, get_data_source: Callable) -> ResolvedReference:
```
KristianKjerstad, line 239: As far as I understand, we should make an effort to substitute all function parameters"reference: string" with "address: Address" in this file.

There are still some functions in this file that take in a reference string, for example
 * _reference_to_reference_items()
 * reference_to_reference_items()
 * _next_reference_part()
 * split_reference()
 
 
 since this PR is already very big, it might be better to update these afterwards

KristianKjerstad (changes requested):
nice work!

Quite a few comments.

Since this will break a lot of frontend functionality, because endpoints have changed, I have created a new tag for dmss: v.1.2.2

we should update all docker-compose files in all repos to use v.1.2.2 instead of latest dmss docker image before merging this PR.

src/common/address.py
```diff
@@ -0,0 +1,48 @@
...
+    def __init__(self, path: str | None, data_source: str, protocol: str = Protocols.DMSS.value):
+        if protocol and protocol != Protocols.DMSS.value:
+            # Only support one reference type
+            raise NotImplementedError(f"The protocol '{protocol}' is not supported")
+        self.protocol = protocol
+        self.data_source = data_source
+        self.path = path
+
+    @classmethod
+    def fromabsolute(cls, address: str):
```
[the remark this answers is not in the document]
ingeridhellen: Mh, yes. I was of the impression that class methods had a different naming convention than ordinary methods, but google does not seem to agree. I'll fix!

src/common/utils/resolve_reference.py
ingeridhellen: I agree. I've linked to this message in #389

PR #460, opened 2023-06-27
refactor: rename reference to path

ingeridhellen:

Rename reference to path inside resolve_reference

Reference refers to a link or storage reference. To avoid confusion, we shouldn't also use reference about what is instead an address path (ie $1.attribute1.attribute2 or [$1, attribute1, attribute2]).

Refs #389


KristianKjerstad (approved):
👍

PR #463, opened 2023-06-29
fix: bug with create entity

### LINKED ISSUE #462: fix: bug with dmss create entity endpoint
reproduce bug:

in the signal app, open case1 and try to add a job.

(this will send a POST request to http://localhost:5000/api/entity with payload: type: "dmss://WorkflowDS/Blueprints/Job")

The job entity that will be cretaed is:

```json
{
    "type": "dmss://WorkflowDS/Blueprints/Job",
    "status": "Not registered",
    "applicationInput": {},
    "result": {},
    "runner": {}
}
```

Correct behavior is that applicationInput, result and runner are not included since they are optional attributes in the Job blueprint. Empty objects are not valid according to the blueprint.

KristianKjerstad:
Before, create entity would create empty dicts for optional attributes. Example:
 POST request to http://localhost:5000/api/entity with payload: type: "dmss://WorkflowDS/Blueprints/Job"

Will create
```json
{
    "type": "dmss://WorkflowDS/Blueprints/Job",
    "status": "Not registered",
    "applicationInput": {},
    "result": {},
    "runner": {}
}
```

after the bug is fixed, optional attributes are not included in the created entity. However, they will be included IF the optional attribute has defined a default value.

Created entity after fix:
```json
{
    "type": "dmss://WorkflowDS/Blueprints/Job",
    "status": "Not registered"
}
```

closes #462

### DIFF
--- a/src/common/utils/create_entity.py
+++ b/src/common/utils/create_entity.py
@@ -71,7 +71,7 @@ class CreateEntity:
 
     # add all non optional attributes with default value.
     # type is inserted based on the parent attributes type, or the initial type for root entity.
-    def _get_entity(self, blueprint: Blueprint, entity):
+    def _get_entity(self, blueprint: Blueprint, entity: dict):
         for attr in blueprint.attributes:
             if attr.attribute_type == BuiltinDataTypes.BINARY.value:
                 continue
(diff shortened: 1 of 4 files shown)


eoaksnes (approved):
Nice

PR #464, opened 2023-06-29
fix: package content should be storage non-contained

eoaksnes:

*  Update the Package blueprint to set the content attribute to contained true (model contained).
* Rerences with reference type storage are not resolved by default. 
  * depth is used to control if references (storage and link) should be resolved (it does not affect contained attributes)
    * depth=0 means that if the address is pointing to a reference it will be returned directly without being resolved first.
    * depth=1 means that if the address is pointing to a reference it will be resolved before returned. Child references will not be resolved.
    * depth=2 means that the entity's direct child references will also be returned.

> BREAKING CHANGE: GET endpoint for documents  - resolve_links is removed

### DIFF
--- a/src/common/utils/get_resolved_document_by_id.py
+++ b/src/common/utils/get_resolved_document_by_id.py
@@ -8,13 +8,7 @@ from storage.data_source_class import DataSource
 
 
 def resolve_reference_list(
-    values: list,
-    document_repository: DataSource,
-    get_data_source,
-    current_id,
-    depth: int = 1,
-    depth_count: int = 0,
-    resolve_references: bool = False,
+    values: list, document_repository: DataSource, get_data_source, current_id, depth: int = 1, depth_count: int = 0
 ) -> list:
     if not values:  # Return an empty list
         return values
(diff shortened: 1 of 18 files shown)


ingeridhellen (changes requested):
The `resolve_references` flag does not add functionality not already present in the `depth`input. I suggest removing `resolve_references`

src/common/utils/get_resolved_document_by_id.py
```diff
@@ -23,27 +17,21 @@ def resolve_reference_list(
...
-            resolve_reference_list(
-                value, document_repository, get_data_source, current_id, depth, depth_count, resolve_links
-            )
+            resolve_reference_list(value, document_repository, get_data_source, current_id, depth, depth_count)
             for value in values
         ]
 
     if is_reference(value_sample):
-        if resolve_links or not is_link(value_sample):
+        if depth_count <= depth:
```
ingeridhellen, line 25: This one is always true. Checking depth vs depth_count is done by resolve_references_in_entity, and since that check must pass before this function can be called, the check is always true.

src/common/utils/get_resolved_document_by_id.py
```diff
@@ -90,14 +76,12 @@ def resolve_references_in_entity(
     current_id: str | None,
     depth: int = 1,
     depth_count: int = 0,
```
ingeridhellen, line 78: I don't see the point of all these default values. Removing them is perhaps not in the scope of this PR though

src/tests/bdd/document/get_entity_with_optional_attributes.feature
```diff
@@ -113,7 +113,7 @@ Feature: Get document
     """
 
   Scenario: when fetching entity by path, optional attributes should not be included
-    Given I access the resource url "/api/documents/test-source-name/TestData/example/?depth=1&resolve_links=true"
+    Given I access the resource url "/api/documents/test-source-name/TestData/example/?depth=1&resolve_references=true"
```
ingeridhellen, line 116:
```suggestion
    Given I access the resource url "/api/documents/test-source-name/TestData/example/?depth=1"
```

src/tests/bdd/document/remove_2.feature
```diff
@@ -61,7 +61,7 @@ Feature: Explorer - Remove by path
     Given i access the resource url "/api/documents/data-source-name/blueprints/sub_package_1"
     When i make a "DELETE" request
     Then the response status should be "OK"
-    Given I access the resource url "/api/documents/data-source-name/$1?resolve_links=True"
+    Given I access the resource url "/api/documents/data-source-name/$1?resolve_references=True"
```
ingeridhellen, line 64:
```suggestion
    Given I access the resource url "/api/documents/data-source-name/$1"
```

ingeridhellen (approved):

PR #467, opened 2023-07-04
fix: document update bug

KristianKjerstad:
Update dmss to fix bug with document update. 

the saving of complex attributes when type is "object" did not work correctly. 
example:
1) updating `data-source-name/$7.complexList[0].complexAttribute` with a PUT request (where type of `complexAttribute` is object)
2) GET `data-source-name/$7.complexList[0].complexAttribute` would give error since it did not exist

Also updated tests to make sure that the document are updated correctly.

bug discovered by working on https://github.com/equinor/dm-core-packages/issues/280


ingeridhellen (approved):
Nice work!

Issue #465, opened 2023-06-30
refactor(dmss): improvements


eoaksnes:
- [ ] 1) decide how-to 2) update the add endpoint that assumes that the input data is resolved (it does not currently work if the input data is a combination of real data and references)
- [ ] 1) decide how-to 2) update the update endpoint that assumes that the input data is resolved (it does not currently work if the input data is a combination of real data and references)
- [x] rename resolve_reference() to resolve_address() @KristianKjerstad #473 
- [x] rename resolve_document() to resolve_references() @KristianKjerstad  #472

PR #473, opened 2023-07-12
refactor: rename resolve_reference() from resolve_reference.py to resolve_address

KristianKjerstad:

refactor resolve_reference() name to resolve_address()
#465


src/common/utils/resolve_reference.py
```diff
@@ -245,8 +243,12 @@ class ResolvedReference:
     entity: dict | list
 
 
-def resolve_reference(address: Address, get_data_source: Callable) -> ResolvedReference:
-    """Resolve the reference into a document."""
+def resolve_address(address: Address, get_data_source: Callable) -> ResolvedReference:
```
ingeridhellen, line 246: Should we also rename ResolvedReference to ResolvedAddress?

src/common/utils/resolve_reference.py
ingeridhellen, line 1: Perhaps rename the file as well?

ingeridhellen (approved):
Good improvement!

src/common/utils/resolve_reference.py
KristianKjerstad: yes, that sounds good

PR #481, opened 2023-08-04
fix: for instantiate entity feature, optional attributes with a default value are included

### LINKED ISSUE #469: fix: create entity endpoint should include an attribute if it is optional AND have a default value
right now, when creating an entity of type EquallySpacedSignal, 
attributes that are optional AND have a default value are not included.

example: description in blueprint is optional, AND havae default value "signal desc", but is not included in entity that is created

Image

KristianKjerstad:

The instantiate entity endpoint should follow these rules:
- all required attributes, as defined in the blueprint, are included.
  If the required attribute has a default value, that value will be used.
  If not, an 'empty' value will be used. For example empty string,
  an empty list, the number 0, etc.
- optional attributes are not included (also true if default value is provided)

This PR fixes the issue that optional attributes with default value are NOT included in the created entity

Also, a refactor of the create_default_array function - it has been moved to a new file and simplified a bit.

closes #469

### DIFF
--- a/src/domain_classes/dimension.py
+++ b/src/domain_classes/dimension.py
@@ -3,53 +3,31 @@ from common.utils.string_helpers import get_data_type_from_dmt_type
 
 class Dimension:
     def __init__(self, dimensions: str, attribute_type: str):
+        """
+        dimensions: define the structure of the array. For example:
+        - "*" is a one dimensional array of any length.
+        - "*,*" is a two dimensional array of any length.
+        - "2,1" is a 2 dimensional array where inner arrays have length one - for example [[1], [3]]
+
+        type: define what type the values in the array have. Primitive types (bool, int, str, etc)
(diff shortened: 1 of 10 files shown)


src/domain_classes/dimension.py
```diff
@@ -4,52 +4,20 @@
...
-    def is_matrix(self):
+    def is_matrix(self) -> bool:
         return len(self.dimensions) > 1
 
     # If the inner most dimension is "*", the Dimension is unfixed
-    def is_unfixed(self):
+    def is_unfixed(self) -> bool:
         return self.dimensions[-1] == "*"
 
-    def create_default_array(self, blueprint_provider, create_entity_class):
```
KristianKjerstad, line 20: function has been moved to the common/utils/arrays.py file

KristianKjerstad:
wait for feedback from @l3abak if the rules for instantiate are correct

PR #486, opened 2023-08-07
refactor: rename insert_reference to update_reference

KristianKjerstad:
also make the function more robust

the old insert_reference() does not work like one expects.
It is actually an update since it is assumed that the attribute where we want to add the reference already exists.

### DIFF
--- a/src/tests/unit/test_reference.py
+++ b/src/tests/unit/test_reference.py
@@ -11,7 +11,7 @@ from tests.unit.mock_utils import get_mock_document_service
 
 
 class ReferenceTestCase(unittest.TestCase):
-    def test_insert_reference(self):
+    def test_update_reference(self):
         repository = mock.Mock()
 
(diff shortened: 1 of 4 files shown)


src/tests/unit/test_reference.py
```diff
@@ -20,7 +20,11 @@ def test_insert_reference(self):
                 "name": "Parent",
                 "description": "",
                 "type": "uncontained_blueprint",
-                "uncontained_in_every_way": {},
```
KristianKjerstad, line 23: having an empty dict here is not valid according to the `uncontained_blueprint`

KristianKjerstad:
decided that the update_reference endpoint can be removed. We can use update document endpoint instead for the same purpose.

must change tests to account for this.

also, make sure to properly document when update document and add document should be used.

KristianKjerstad:
replaced by #490

PR #488, opened 2023-08-08
refactor: create_default_array

KristianKjerstad:

- a refactor of the create_default_array function - it has been moved to a new file and simplified a bit.
- refactor of tests
- bdd test for instantiate entity that was missing

### DIFF
--- a/src/common/utils/arrays.py
+++ b/src/common/utils/arrays.py
@@ -11,6 +11,22 @@ def create_default_array(
     ],  # TODO use Type[CreateEntity]. can't use it now due to circular import. Needs refactoring.
     default_array_value: list[Any] = None,
 ) -> list:
+    """
+    Create an array with default values with a given dimension.
+
+    - Dimension.dimensions = ["2,1"] represents a 2 dimensional array where inner arrays have length one - for example [[1], [3]]
+    - Dimension.dimensions = ["*"] represents a one dimensional array of any length.
+
+    Hard coded default values for different types (type is specified in the Dimension class):
+    - integer: 0
+    - number: 0.00
+    - boolean: False
+    - string: ""
+
+    For complex types (For example dmss://system/SIMOS/Package), default value is found from the 'create_entity_class'.
+
+    For dimension ["*"], default_array_value can be used to create an array with a given default value.
+    """
     dimensions = dimension.dimensions
     if dimensions == [""]:
         raise Exception("This attribute is not an array!")
--- a/src/common/utils/arrays.py
+++ b/src/common/utils/arrays.py
@@ -1,0 +1,74 @@
+from typing import Any, Callable, Type
+
+from domain_classes.blueprint_attribute import BlueprintAttribute
+from domain_classes.dimension import Dimension
+
+
+def create_default_array(
+    dimension: Dimension,
+    blueprint_provider: Callable,
+    create_entity_class: Type[
+        Any
+    ],  # TODO use Type[CreateEntity]. can't use it now due to circular import. Needs refactoring.
+    blueprint_attribute: BlueprintAttribute,
+) -> list:
+    dimensions = dimension.dimensions
+    if dimensions == [""]:
+        raise Exception("This attribute is not an array!")
+    if len(dimensions) == 1:
+        if dimensions[0] == "*":
+            return blueprint_attribute.default if blueprint_attribute.default is not None else []
+
+        # TODO use default value from blueprint if it exists for cases other than dimensions[0] == "*"
+        # Return a list initialized with default values for the size of the array.
+        if dimension.type is int:
+            return [0 for n in range(int(dimensions[0]))]
+        if dimension.type is float:
+            return [0.00 for n in range(int(dimensions[0]))]
+        if dimension.type is str:
+            return ["" for n in range(int(dimensions[0]))]
+        if dimension.type is bool:
+            return [False for n in range(int(dimensions[0]))]
+        else:
+            # For fixed complex types, create the entity with default values. Set name from list index.
+            return [create_entity_class(blueprint_provider, dimension.type).entity for n in range(int(dimensions[0]))]
+
+    if dimensions[0] == "*":
+        # If the size of the rank is "*" we only create one nested list.
+        nested_list = [
+            create_default_array(
+                Dimension(remove_first_and_join(dimensions), dimension.type),
+                blueprint_provider,
+                create_entity_class,
+                blueprint_attribute,
+            )
+        ]
+    else:
+        # If the size of the rank in NOT "*", we expect an Integer, and create n number of nested lists.
+        nested_list = [
+            create_default_array(
+                Dimension(remove_first_and_join(dimensions), dimension.type),
+                blueprint_provider,
+                create_entity_class,
+                blueprint_attribute,
+            )
+            for n in range(int(dimensions[0]))
+        ]
+    return nested_list
+
+
+def remove_first_and_join(input_list: list[str]) -> str:
+    """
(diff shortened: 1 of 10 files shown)


src/common/utils/arrays.py
```diff
@@ -0,0 +1,74 @@
...
+        nested_list = [
+            create_default_array(
+                Dimension(remove_first_and_join(dimensions), dimension.type),
+                blueprint_provider,
+                create_entity_class,
+                blueprint_attribute,
+            )
+            for n in range(int(dimensions[0]))
+        ]
+    return nested_list
```
ingeridhellen, line 57: Perhaps refactor to
```suggestion
    return  [
        create_default_array(
            Dimension(remove_first_and_join(dimensions), dimension.type),
            blueprint_provider,
            create_entity_class,
            blueprint_attribute,
        )
        for n in range(1 if dimensions[0] == "*" else int(dimensions[0]))
    ]
```

src/common/utils/arrays.py
```diff
@@ -0,0 +1,74 @@
...
+        Any
+    ],  # TODO use Type[CreateEntity]. can't use it now due to circular import. Needs refactoring.
+    blueprint_attribute: BlueprintAttribute,
+) -> list:
+    dimensions = dimension.dimensions
+    if dimensions == [""]:
+        raise Exception("This attribute is not an array!")
+    if len(dimensions) == 1:
+        if dimensions[0] == "*":
+            return blueprint_attribute.default if blueprint_attribute.default is not None else []
```
ingeridhellen, line 20: This functionality is not tested by the unit tests. Also, I wonder, shouldn't we always use the default value no matter what the dimensions are?

src/common/utils/arrays.py
```diff
@@ -0,0 +1,74 @@
...
+from domain_classes.dimension import Dimension
+
+
+def create_default_array(
+    dimension: Dimension,
+    blueprint_provider: Callable,
+    create_entity_class: Type[
+        Any
+    ],  # TODO use Type[CreateEntity]. can't use it now due to circular import. Needs refactoring.
+    blueprint_attribute: BlueprintAttribute,
```
ingeridhellen, line 13: This parameter confused me at first, cause I didn't understand why you would need it. I see now that you only need blueprint_attribute.default. So perhaps instead do
```suggestion
    default: any,
```
That would be more intuitive, I think, but up to you

src/domain_classes/dimension.py
```diff
@@ -3,53 +3,31 @@
 
 class Dimension:
     def __init__(self, dimensions: str, attribute_type: str):
+        """
+        dimensions: define the structure of the array. For example:
+        - "*" is a one dimensional array of any length.
+        - "*,*" is a two dimensional array of any length.
+        - "2,1" is a 2 dimensional array where inner arrays have length one - for example [[1], [3]]
+
+        type: define what type the values in the array have. Primitive types (bool, int, str, etc)
```
ingeridhellen, line 12:
```suggestion
        attribute_type: define what type the values in the array have. Primitive types (bool, int, str, etc)
```
Also, I miss an example of valid input. Should I pass in "int" or "integer"?

src/domain_classes/dimension.py
```diff
@@ -3,53 +3,31 @@
...
-        self.type: str = get_data_type_from_dmt_type(attribute_type)
+        self.type: type[bool] | type[int] | type[float] | type[str] | str = get_data_type_from_dmt_type(attribute_type)
         self.value = None
 
-    def is_array(self):
+    def is_array(self) -> bool:
         return self.dimensions != [""]
 
-    def is_matrix(self):
+    # TODO remove function - it is never used for anything useful
```
ingeridhellen, line 23: Or rewrite to `get_dimensions(self) -> int`?

src/common/utils/create_entity.py
```diff
@@ -47,7 +48,7 @@ def parse_value(attr: BlueprintAttribute, blueprint_provider):
 
         if default_value is None:
             if attr.is_array:
-                return attr.dimensions.create_default_array(blueprint_provider, attr)
+                return attr.dimensions.create_default_array_recursive(blueprint_provider, attr)
```
ingeridhellen, line 51: Is this working? Where is this function declared?

src/common/utils/create_entity.py
```diff
@@ -69,13 +70,22 @@ def is_json(attr: BlueprintAttribute):
             ""
         ]  # type(attr.default) == dict or type(attr.default) == list
 
-    # add all non optional attributes with default value.
     # type is inserted based on the parent attributes type, or the initial type for root entity.
     def _get_entity(self, blueprint: Blueprint, entity: dict):
+        """
```
ingeridhellen, line 75: Great! Could this be moved up to line 26? As a docstring to the class instead of a private function?

src/common/utils/create_entity.py
```diff
@@ -69,13 +70,22 @@ def is_json(attr: BlueprintAttribute):
...
         for attr in blueprint.attributes:
             if attr.attribute_type == BuiltinDataTypes.BINARY.value:
                 continue
-
+            if attr.is_optional:
+                # skip attribute if it is optional
+                continue
             if attr.attribute_type in PRIMITIVES:
                 if not attr.is_optional and attr.name not in entity:
```
ingeridhellen, line 90:
```suggestion
                if attr.name not in entity:
```

src/features/entity/entity_feature.py
```diff
@@ -19,6 +19,13 @@ def instantiate(entity: Entity, user: User = Depends(auth_w_jwt_or_pat)):
     """Create a new entity and return it.
 
     (entity is not saved in DMSS)
+    Rules for instantiation:
+    - all required attributes, as defined in the blueprint, are included.
+      If the required attribute has a default value, that value will be used.
+      If not, an 'empty' value will be used. For example empty string,
+      an empty list, the number 0, etc.
+    - optional attributes with a default value are included
```
ingeridhellen, line 27: This is wrong

src/tests/bdd/entity/instantiate_entity.feature
```diff
@@ -0,0 +1,124 @@
...
+        }
+      ]
+    }
+    """
+    Given there exist document with id "3" in data source "data-source-name"
+    """
+    {
+      "type": "dmss://data-source-name/root_package/Employee",
+      "name": "BigBossMan",
+      "isManager": true
```
ingeridhellen, line 97: He's missing the non-optional managers attribute

ingeridhellen (changes requested):
Good changes! I've been extra difficult on this one, so feel free to dismiss my suggestions if you don't agree with them.

src/domain_classes/dimension.py
KristianKjerstad: ah, right. The primitive types should be written out in full: boolean, integer, string and number. WIll update the docstring.

src/common/utils/arrays.py
KristianKjerstad: Added a simple test for this. 

Yes, Ideally we should use the default value for all dimensions. I added a TODO for that a few lines below, but I can create another issue for this. Should probably not fix it here to avoid increasing the PR size.

src/common/utils/arrays.py
KristianKjerstad: nice!

src/domain_classes/dimension.py
KristianKjerstad: since I didn't see any apparent uses of get_dimensions() (without too much investigation), decided to skip this one.

ingeridhellen (approved):
Nice!

PR #494, opened 2023-08-09
fix: remove insert_reference

### LINKED ISSUE #490: fix: delete insert_reference endpoint
It was decided that the insert_reference endpoint is not needed. We can use document update endpoint instead.

KristianKjerstad:
Document add endpoint can be used instead of the insert reference endpoint.

closes #490

### DIFF
--- a/src/features/document/document_feature.py
+++ b/src/features/document/document_feature.py
@@ -78,6 +78,11 @@ def add_document(
       - Reference to package by id: PROTOCOL://DATA SOURCE/$ID
       - Reference to package by path: PROTOCOL://DATA SOURCE/ROOT PACKAGE/SUB PACKAGE
       The PROTOCOL is optional, and the default is dmss.
+
+    This endpoint can be used for:
+    - Adding elements to a list attribute in an entity.
+    - Adding a new document to a package / data source
+    - Adding an object to an entity (for example filling in an optional, complex attribute)
     """
     return add_document_use_case(
         user=user,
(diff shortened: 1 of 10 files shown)


src/tests/unit/test_reference.py
```diff
@@ -42,90 +44,54 @@ def mock_update(entity: dict, *args, **kwargs):
...
         }
-        document_service.insert_reference(
-            Address("$1.uncontained_in_every_way", "testing"),
-            ReferenceEntity.parse_obj(reference),
+        document_service.update_document(
+            Address("$1.uncontained_in_every_way", "testing"), reference, update_uncontained=False
         )
         assert doc_storage["1"]["uncontained_in_every_way"] == reference
 
-    def test_insert_reference_target_does_not_exist(self):
```
KristianKjerstad, line 51: document update does not have validation for checking if reference target exists

src/features/document/document_feature.py
```diff
@@ -78,6 +78,8 @@ def add_document(
       - Reference to package by id: PROTOCOL://DATA SOURCE/$ID
       - Reference to package by path: PROTOCOL://DATA SOURCE/ROOT PACKAGE/SUB PACKAGE
       The PROTOCOL is optional, and the default is dmss.
+
+    This endpoint can be used to add elements to a list attribute in an entity.
```
ingeridhellen, line 82: In the reference tests you're using it to fill an empty object. Not sure what is correct use of the add endpoint

ingeridhellen (approved):

src/features/document/document_feature.py
KristianKjerstad: I believe it should be both adding elements to list and filling empty object. Can update the docs

Issue #485, opened 2023-08-07
fix: adding optional object expects object to already exist


ingeridhellen:
Found endpoints affected, but it may be more:
- documentAdd @eoaksnes 

If you have a blueprint with an optional attribute myOptionalAttribute, an entity of that blueprint does not have to include myOptionalAttribute at all. However, if you try to add a document or reference to that attribute, you get an error.

- [x] fix bug in add() for document_service.py https://github.com/equinor/data-modelling-storage-service/pull/497

KristianKjerstad:
wait until document service refactor is done (move functionality out to use cases)

PR #497, opened 2023-08-11
fix: bug with document add endpoint for optional attributes

KristianKjerstad:
fix bug in document add endpoint when adding optional attribute that does not exist

https://github.com/equinor/data-modelling-storage-service/issues/485

### DIFF
--- a/src/services/document_service.py
+++ b/src/services/document_service.py
@@ -424,6 +424,7 @@ class DocumentService:
             #      the parent of entity referenced by 'address'.
             #   2) the address is wrong. In that case, raise Exception.
 
+            # It is assumed that address.path is to an attribute, e.g. dataSource/package/document.attribute
             split_address_path: list[str] = address.path.rsplit(".", 1)
 
             if len(split_address_path) <= 1 and split_address_path[0] == address.path:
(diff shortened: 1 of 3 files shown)


src/services/document_service.py
```diff
@@ -413,10 +413,67 @@ def add(
...
+
+            split_address_path: list[str] = address.path.rsplit(".", 1)
+
+            if len(split_address_path) <= 1 and split_address_path[0] == address.path:
+                # Raising NotFoundException, since the get_document() did not find the document
+                # with address=address.path in the above try except statement
+                raise NotFoundException(f"Could not find document {address}")
+
+            last_attribute_in_address: str = split_address_path.pop()
+            parent_address_as_string: str = ".".join(split_address_path)
```
pbullhove, line 435:
```suggestion
            parent_address_as_string, last_attribute_in_address = split_address
```
I think you don't have to do this pop and join anymore since you did rsplit() and not split() on line 427, so should be enough to do the unpacking since the array will only be two elements long at most.

pbullhove (approved):
approved with a suggestion

src/services/document_service.py
KristianKjerstad: nice!

PR #498, opened 2023-08-14
refactor: rename files to be equal to their use_case_functions

pbullhove:

Renames two files, and modifies references, such that they are according to convention. The file should be the same as the function inside it. 

remove.py -> remove_use_case.py

add_document_to_path_use_case.py -> add_document_use_case.py

Convention is key bruh. 

Nah.


KristianKjerstad (approved):
👍

PR #506, opened 2023-08-17
ci: add release please

ingeridhellen:

Adds release please, with automatic changelog and update of version in app and pyproject.toml

Note that I've been able to verify that the release-please workflow works (https://github.com/equinor/data-modelling-storage-service/pull/505), but not the publish workflows, so there might be bugs there.

Easier release


KristianKjerstad (approved):
looks good!

PR #508, opened 2023-08-17
refactor: Rename string classes to convention

pbullhove:

Custom string types should have the CamelCase notation in order to not be confused with variables. 

Convention is key. Without it we are just swimming in spaghetti.

### DIFF
--- a/src/features/entity/entity_feature.py
+++ b/src/features/entity/entity_feature.py
@@ -4,7 +4,7 @@ from fastapi.responses import JSONResponse
 from authentication.authentication import auth_w_jwt_or_pat
 from authentication.models import User
 from common.responses import create_response, responses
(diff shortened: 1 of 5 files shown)


src/features/entity/entity_feature.py
```diff
@@ -16,9 +16,17 @@
...
 @create_response(JSONResponse)
 def instantiate(entity: Entity, user: User = Depends(auth_w_jwt_or_pat)):
-    """Create a new entity and return it.
+    """Returns a default entity of specified type. This entity is not stored in the database.
+
+
+    For storing an entity in the database, see the Post Document endpoint.
+
+    Args:
+    -
```
KristianKjerstad, line 25: can add entity to the argument list

PR #510, opened 2023-08-18
refactor: move add document logic to use case

eoaksnes:

* Moves the add document from the document service to the `add_document_use case`.

* Document service is too big, and getting an overview of the code is therefore more complicated than it should. Isolate all document add functionality to the add document feature (use case). 

> NOTE: this is the first PR of several that will refactor the document service.

### DIFF
--- a/src/features/document/use_cases/add_document_use_case.py
+++ b/src/features/document/use_cases/add_document_use_case.py
@@ -1,28 +1,220 @@
-from typing import List, Optional
+from typing import BinaryIO, List, Optional
 
 from fastapi import UploadFile
 
-from authentication.models import User
 from common.address import Address
+from common.exceptions import (
+    BadRequestException,
+    NotFoundException,
+    ValidationException,
+)
+from common.tree_node_serializer import tree_node_from_dict
+from common.utils.merge_entity_and_files import merge_entity_and_files
+from common.utils.validators import validate_entity, validate_entity_against_self
+from domain_classes.blueprint_attribute import BlueprintAttribute
+from domain_classes.tree_node import ListNode, Node
+from enums import SIMOS
+from restful.request_types.shared import Entity
 from services.document_service import DocumentService
-from storage.internal.data_source_repository import get_data_source
+
+
+def _add_document_to_data_source(
+    data_source_id: str, document: dict, update_uncontained: Optional[bool], document_service: DocumentService
+) -> dict:
+    """Add the document to an existing data source.
+
+    Args:
+       data_source_id: The data source ID
+       document: The entity to be added
+       update_uncontained: Whether to update uncontained children (deprecated)
+       document_service: The document service
+
+    Returns:
+       A dict that contains the ID of the added document.
+    """
+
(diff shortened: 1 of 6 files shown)


src/features/document/use_cases/add_document_use_case.py
```diff
@@ -1,28 +1,220 @@
...
-    address: str,
+    address: Address,
+    document_service: DocumentService,
     files: Optional[List[UploadFile]] = None,
     update_uncontained: Optional[bool] = False,
-    repository_provider=get_data_source,
-):
-    document_service = DocumentService(repository_provider=repository_provider, user=user)
+) -> dict:
+    """Add document to a data source or existing entity.
```
KristianKjerstad, line 197: the use case is also used to add items to list
```suggestion
    """Add document to a data source or existing entity. Can also be used to add (complex) items to a list. 
```

src/features/document/use_cases/add_document_use_case.py
```diff
@@ -1,28 +1,220 @@
...
+
+    Returns:
+        A dict that contains the ID of the added document.
+    """
+    validate_entity_against_self(document, document_service.get_blueprint)
+
+    if not address.path:
+        return _add_document_to_data_source(address.data_source, document, update_uncontained, document_service)
+
+    return _add_document_to_entity(
```
KristianKjerstad, line 214:
```suggestion
    return _add_document_to_entity_or_list(
```

src/features/document/use_cases/add_document_use_case.py
```diff
@@ -1,28 +1,220 @@
...
-):
-    document_service = DocumentService(repository_provider=repository_provider, user=user)
+) -> dict:
+    """Add document to a data source or existing entity.
 
-    document = document_service.add(
-        address=Address.from_absolute(address),
+    Args:
+        document: The entity to be added
+        address: Reference to a package or a data source
```
KristianKjerstad, line 201:
```suggestion
        address: Reference to a package, attribute inside an entity (either a list or a complex attribute) or a data source
```

src/features/document/use_cases/add_document_use_case.py
```diff
@@ -1,28 +1,220 @@
...
+    address: Address,
+    document: dict,
+    files: dict[str, BinaryIO] | None,
+    update_uncontained: Optional[bool],
+    document_service: DocumentService,
+) -> dict:
+    """Add the document to an existing entity.
+
+    Args:
+       address: Reference to an existing entity
```
KristianKjerstad, line 71:
```suggestion
       address: Reference to an existing entity or to an attribute (complex or list attribute) inside an entity.
```

KristianKjerstad (commented):
For this structure to make sense it looks like we have to remove the entire document service and move all that logic into use cases. Is that the plan?

eoaksnes:
> For this structure to make sense it looks like we have to remove the entire document service and move all that logic into use cases. Is that the plan?

Not everything will be removed, the get and save will still be part of the document service (or in the shape of another component). Since all CRUD  operations will be using get and save.

pbullhove (approved):

PR #520, opened 2023-08-21
refactor: Rename of a method to fix private method import issue

pbullhove:

1. Renamed get_access_control to get_lookup because that is what is really is returning.  
2. Changed reference to use get_lookup instead of using the private method _lookup ( ) 

Convention and clear naming is key


KristianKjerstad (approved):
🎉

PR #521, opened 2023-08-21
refactor: renaming to be more clear

pbullhove:

Renames some things to be more clear
1.  ACL -> AccessControlList
2.  access_control( ) -> assert_user_has_access( )
3. DEFAULT_ACT -> DEFAULT_ACCESS_CONTROL_LIST

Understanding this as a new guy is really hard, so therefore it is better to write out the names of code. 
Code should be more readable as natural language, and not filled with abbreviations and vague method names.


eoaksnes (approved):
Good refactoring makes the code easier to read! 👍

PR #524, opened 2023-08-21
refactor: refactored search-use-case

pbullhove:

Refactors the search-use-case method which was messy and had code duplications. 
Not it is hopefully more clear and straight-forward. 

1. Removed method output variable in python which is a true code smell. 
2. Removed else-statement, which kinda clutters the code 
3. Removed duplicate for loops with calls to same function, now only one. 

See why here: 
https://www.youtube.com/watch?v=eUhWKe1ZqKY

### DIFF
--- a/src/features/search/use_cases/search_use_case.py
+++ b/src/features/search/use_cases/search_use_case.py
@@ -15,30 +15,23 @@ class SearchRequest(DataSourceList):
     dotted_attribute_path: str
 
 
-def get_search_result(output_dict: dict, data_source_id: str, document_service: DocumentService, req: SearchRequest):
-    results = document_service.search(
-        data_source_id=data_source_id,
-        search_data=deepcopy(req.data),
-        dotted_attribute_path=req.dotted_attribute_path,
-    )
-    return output_dict.update(results)
-
-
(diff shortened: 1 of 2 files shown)


src/features/search/use_cases/search_use_case.py
```diff
@@ -15,30 +15,23 @@ class SearchRequest(DataSourceList):
...
     document_service: DocumentService = DocumentService(repository_provider=repository_provider, user=user)
-    all_data_sources = DataSourceRepository(user).list()
+    all_data_source_ids = [ds["id"] for ds in DataSourceRepository(user).list()]
+    search_data_sources = all_data_source_ids
 
-    search_results: dict = {}
+    if request.data_sources:
+        # If user has specified any data sources, check that they exist, then select only through them for search.
+        if invalid_search_data_sources := set(request.data_sources) - set(all_data_source_ids):
+            raise BadRequestException(f"Data source {invalid_search_data_sources.pop()} not found")
```
ingeridhellen, line 26: If there are more than one invalid data source, it might be better to print the ids of all of the invalid ones, not just the first one.

ingeridhellen (approved):
Much better!

PR #528, opened 2023-08-23
Fix/delete required attribute

collinlokken:
Introduces a check that will not allow deletion of required attributes. This also changes some tests, as a couple of test tries to delete required attributes.

#516

### DIFF
--- a/src/tests/unit/test_remove.py
+++ b/src/tests/unit/test_remove.py
@@ -3,6 +3,7 @@ from copy import deepcopy
 from unittest import mock
 
 from common.address import Address
+from common.exceptions import ValidationException
 from common.utils.data_structure.compare import get_and_print_diff
 from enums import REFERENCE_TYPES, SIMOS
 from tests.unit.mock_utils import get_mock_document_service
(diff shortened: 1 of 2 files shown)


src/tests/unit/test_remove.py
```diff
@@ -38,21 +39,39 @@ def test_remove_document(self):
...
+                "type": "dmss://testing/this_blueprint_does_not_exist",
+            }
         }
 
-        class NoBlueprints:
-            def get_blueprint(self, type):
-                raise FileNotFoundError
+        # class NoBlueprints:
+        #    def get_blueprint(self, type):
+        #        raise FileNotFoundError
```
KristianKjerstad, line 69:
```suggestion
```
can be deleted

src/tests/unit/test_remove.py
```diff
@@ -84,9 +103,10 @@ def test_remove_child_dict(self):
             "2": {"name": "Nested", "description": "", "type": "basic_blueprint"},
         }
 
-        self.document_service.remove(Address("$1.nested", "testing"))
-        assert self.storage["1"].get("nested") is None
-        assert self.storage.get("2") is None
+        self.assertRaises(ValidationException, self.document_service.remove, Address("$1.nested", "testing", "dmss"))
+        # TODO Test fails due to modelling error in blueprint. Cannot delete required attribute
```
KristianKjerstad, line 107: would be nice to also add a new issue to the "incoming" tab in the board to fix modeling errors in the blueprints used in this test_remove.py also 👍

KristianKjerstad (approved):
nice!

src/tests/unit/test_remove.py
collinlokken: Ok, I will add it to the board!

PR #532, opened 2023-08-23
refactor: changed parameter from dict to str

pbullhove:

Changes parameter in instansiate to str instead of dict. 
Also changes some names in order to make it more readable.


KristianKjerstad (approved):
good changes! 
(There is a chance this will break things in dm-core-packages if latest dmss docker image is still used there)

PR #537, opened 2023-08-24
Fix/delete required attribute tests

collinlokken:


eoaksnes (approved):
Nice, good if you remove the comments (old code)

PR #540, opened 2023-08-24
feat: endpoint for checking if document exist

### LINKED ISSUE #529: feat: add endpoint to check that an address exist or not
- [ ] return 200 and 404 or just use 200 and reply response

collinlokken:
Introduces an endpoint for checking if a document exists. The endpoint returns a boolean based on a document address, and a 200 OK for both cases. 

Gives the ability to check if a document exists without creating error logs.

Closes #529

### DIFF
--- a/src/features/document/document_feature.py
+++ b/src/features/document/document_feature.py
@@ -15,6 +15,7 @@ from storage.internal.data_source_repository import get_data_source
 
 from .use_cases.add_document_use_case import add_document_use_case
 from .use_cases.add_raw_use_case import add_raw_use_case
+from .use_cases.check_exsistence_use_case import check_existence_use_case
 from .use_cases.get_document_use_case import get_document_use_case
 from .use_cases.remove_use_case import remove_use_case
 from .use_cases.update_document_use_case import update_document_use_case
@@ -174,3 +175,18 @@ def remove(address: str, user: User = Depends(auth_w_jwt_or_pat)):
     - str: "OK" (200)
     """
     return remove_use_case(user=user, address=address)
+
+
+@router.get("-existence/{address:path}", operation_id="document_check", response_model=bool, responses=responses)
+@create_response(JSONResponse)
+def check_existence(address: str, user: User = Depends(auth_w_jwt_or_pat)):
+    """Checks if an entity exists, given an address.
+
+    Args:
+    - Address
+
+    Returns:
+    - bool: 'true' if the address points to an existing document, else 'false'.
+    """
+    document_service = DocumentService(repository_provider=get_data_source, user=user)
+    return check_existence_use_case(address=Address.from_absolute(address), document_service=document_service)
--- a/src/tests/bdd/document/check_existence.feature
+++ b/src/tests/bdd/document/check_existence.feature
@@ -1,0 +1,34 @@
+# Created by christopher.lokken at 24/08/2023
+Feature: Check the existence of documents
+
+  Background: There are data sources in the system
+    Given the system data source and SIMOS core package are available
+    Given there are basic data sources with repositories
+      |   name  |
+      | test-DS |
+    Given there exist document with id "1" in data source "test-DS"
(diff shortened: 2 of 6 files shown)


src/tests/unit/test_remove.py
```diff
@@ -20,7 +19,7 @@ def setUp(self):
         self.document_service = get_mock_document_service(lambda x, y: self.repository)
 
     def mock_get(self, document_id: str):
-        return deepcopy(self.storage[document_id])
```
collinlokken, line 23: automatically removed by pre-commit. I thought it was a bit strange, but...

src/features/document/document_feature.py
```diff
@@ -174,3 +175,18 @@ def remove(address: str, user: User = Depends(auth_w_jwt_or_pat)):
     - str: "OK" (200)
     """
     return remove_use_case(user=user, address=address)
+
+
+@router.get("-existence/{address:path}", operation_id="document_check", response_model=bool, responses=responses)
```
KristianKjerstad, line 180: is it possible to define the endpoint to be /address/exists? Not sure if fast api is able to interpret the url parameter address in this case. But the url /documents/address/exists is a bit easier to read than /documents-existence/address IMO
```suggestion
@router.get("{address:path}/exists", operation_id="document_check", response_model=bool, responses=responses)
```

src/features/document/document_feature.py
```diff
@@ -174,3 +175,18 @@ def remove(address: str, user: User = Depends(auth_w_jwt_or_pat)):
...
+@create_response(JSONResponse)
+def check_existence(address: str, user: User = Depends(auth_w_jwt_or_pat)):
+    """Checks if a document exists, given an address.
```
KristianKjerstad, line 183:
```suggestion
    """Checks if an entity exists, given an address.
```

when address is datasource/path/car.engine, the address points to an entity and not a document. This is a definition thing, but all documents are actually entities (but not all entities are documents).

Documents are basically what is stored in mongo db as a separate "entry"

src/tests/bdd/document/check_existence.feature
```diff
@@ -0,0 +1,34 @@
...
+      "content": [],
+      "isRoot": true
+    }
+    """
+
+  Scenario: Check the existence of non-existing document
+    Given I access the resource url "/api/documents-existence/test-DS/$1"
+    When I make a "GET" request
+    Then the response status should be "OK"
+    And the response should contain
```
KristianKjerstad, line 24:
```suggestion
    And the response should be
```
better to use the "... should be" function if we know the exact result the endpoint should return

src/tests/bdd/document/check_existence.feature
```diff
@@ -0,0 +1,34 @@
...
+    """
+    true
+    """
+    Given I access the resource url "/api/documents-existence/test-DS/$2"
+    When I make a "GET" request
+    Then the response status should be "OK"
+    And the response should contain
```
KristianKjerstad, line 31:
```suggestion
    And the response should be
```
better to use the "... should be" function if we know the exact result the endpoint should return

src/tests/unit/test_check_existence.py
```diff
@@ -0,0 +1,40 @@
...
+    check_existence_use_case,
+)
+from tests.unit.mock_utils import get_mock_document_service
+
+
+class CheckExistenceTestCase(unittest.TestCase):
+    def setUp(self) -> None:
+        self.storage = {
+            "1": {
+                "name": "Test document",
```
KristianKjerstad, line 16:
```suggestion
                "name": "Test_document",
```
We have defined a rule that name should not contain spaces (see EntityName class in shared.py). Not sure if we have added validation for that...

KristianKjerstad (commented):
Nice!

A few suggestions for improvement.

KristianKjerstad (approved):
🎉

PR #546, opened 2023-08-30
fix: add job to case

collinlokken:
Fixes so that it is possible to add a job to case

Jobs will currently not be added to a case. A job node is created in the backend and saved, but is not appended to the parent Stody.Cases[0] document.

https://github.com/equinor/dm-core-packages/issues/437

### DIFF
--- a/src/features/document/use_cases/add_document_use_case.py
+++ b/src/features/document/use_cases/add_document_use_case.py
@@ -118,27 +118,42 @@ def _add_document_to_entity_or_list(
                 f"Could not find attribute {last_attribute_in_address} in blueprint for {parent_blueprint.name}"
             )
 
-        parent_document = parent_node.entity
         attribute_to_update = [
             blueprint_attribute
             for blueprint_attribute in parent_blueprint.attributes
             if blueprint_attribute.name == last_attribute_in_address
         ][0]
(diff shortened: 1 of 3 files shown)


src/features/document/use_cases/add_document_use_case.py
```diff
@@ -118,27 +118,43 @@ def _add_document_to_entity_or_list(
...
-            parent_document[last_attribute_in_address] = document
 
-        target = parent_node
+        node_attribute = BlueprintAttribute(name=attribute_to_update.name, attribute_type=entity.type)
         new_node = tree_node_from_dict(
-            {**parent_document},
+            {**document},
             blueprint_provider=document_service.get_blueprint,
-            node_attribute=BlueprintAttribute(name=target.attribute.name, attribute_type=entity.type),
+            node_attribute=node_attribute,
```
eoaksnes, line 131: Suggestion for change:

```
        new_node = tree_node_from_dict(
            {**document},
            blueprint_provider=document_service.get_blueprint,
            node_attribute=BlueprintAttribute(name=attribute_to_update.name, attribute_type=entity.type),
            recipe_provider=document_service.get_storage_recipes,
        )
```

eoaksnes (approved):

PR #552, opened 2023-09-05
refactor: clean up some unit tests

pbullhove:

1. Moved some tests under _test_data_structure_ folder according to what they are testing 
2. Renamed "is_same" to "has_key_value_pair" which is what it actually is testing. 
3. Rename _reference_ --> _path_ to comply with the methods they are testin. 
4. Moved test method for '_next_path_part' into another folder because unrelated to what it was in before. 
5. Changed some test methods names


ingeridhellen (approved):
Nice!

PR #553, opened 2023-09-05
Feat: error handling in 'has_key_value_pair' method

pbullhove:

Merge #552 first

Example 
`
input: 
    target {"a": "b"}
    content {"x": "y"}
` -> throws
changed to -> `returns false`

The way it was implemented was that 
- if the key in `content` did not exist, the method throws! 

From the naming of the method, and the use, I propose that it should return `false` when the key in `content` does not exist in `target`. 

should return false instead of throw. 

The method `has_key_value_pair()` is a method that is used only once, in `get_child()`, se below, and this implementation should not break the usage of the method

image

### DIFF
--- a/src/tests/unit/test_common/test_utils/test_data_structure/test_has_key_value_pairs.py
+++ b/src/tests/unit/test_common/test_utils/test_data_structure/test_has_key_value_pairs.py
@@ -10,7 +10,14 @@ class HasKeyValuePairsTestCase(unittest.TestCase):
     def test_target_contains_key_value_pair_returns_true(self):
         assert has_key_value_pairs(self.target, {"a": "1"})
         assert has_key_value_pairs(self.target, {"b": "2"})
+        assert has_key_value_pairs(self.target, {"b": "2", "a": "1"})
 
     def test_target_not_contains_key_value_pair_returns_false(self):
         assert has_key_value_pairs(self.target, {"a": "2"}) is False
         assert has_key_value_pairs(self.target, {"b": "1"}) is False
+        assert has_key_value_pairs(self.target, {"a": "1", "b": "1000"}) is False
+
+    def test_target_not_contains_key_returns_false(self):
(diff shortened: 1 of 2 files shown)


src/tests/unit/test_common/test_utils/test_data_structure/test_has_key_value_pairs.py
```diff
@@ -14,3 +14,8 @@ def test_target_contains_key_value_pair_returns_true(self):
     def test_target_not_contains_key_value_pair_returns_false(self):
         assert has_key_value_pairs(self.target, {"a": "2"}) is False
         assert has_key_value_pairs(self.target, {"b": "1"}) is False
+
+    def test_target_not_contains_key_returns_false(self):
+        assert has_key_value_pairs(self.target, {"x": "y"}) is False
+        assert has_key_value_pairs(self.target, {1: {"x": "y"}}) is False
+        assert has_key_value_pairs(self.target, {1: {"a": "1"}}) is False
```
ingeridhellen, line 21: Could you try adding a test where `content=={"a": "1", "b": "1000"}`? I think the function might return true then, when it really should return false

ingeridhellen (changes requested):

src/tests/unit/test_common/test_utils/test_data_structure/test_has_key_value_pairs.py
pbullhove: Added test with the content you specified. Pushed to ref now. 
Unit tests still run green!

pbullhove: I see! I misunderstood. Thanks for clarifying. Will fix it now

ingeridhellen (approved):

eoaksnes (approved):
Nice cleanup 👍

PR #554, opened 2023-09-05
refactor: remove unused methods groupBy, toPath, get_by_dot

pbullhove:

Removed 3 methods that had no uses. As well as removing their tests. 
This is done to keep the code-base small and clean. If they are required later, we can find them easily in the git history. 

❌ `to_path` (a method that turned dotted string notation into a list of strings. ) 
                - `example: "parent.child.grandChild" -> ["parent", "child", "grandChild"]`
❌ `get_by_dot` (a method for getting a nested element in a dict using a dotted string path, like attribute.attribute.attribute) 
❌  `group_by`  (a method for grouping elements in a dict based on common elements) 

Keeping the codebase clean


ingeridhellen (approved):
🎉

PR #559, opened 2023-09-07
feat: extend testing coverage of scrypt method to be complete

pbullhove:

Was digging through unit tests, and found hasing testing to be incomplete. 
- It did only test that equal values with the same salt returned the same hash. 

1. I added the negative test, that for different salts and values, the hash should _not_ be equal. 
3. Test that the output of the salting has the same length, for different input
4. Test that the output does not contain the values or the salt, so that it _actually_ is hashed.


ingeridhellen (approved):
Nice!

PR #561, opened 2023-09-08
Medium - Simpler blueprint provider mocker class

pbullhove:

The BlueprintProvider mock was hard-coded per BlueprintName. 
The logic is now moved into being that the fileName is the same as the type. 
This makes it very easy to find the one you are looking for, and it makes the Mock very light-weight. 

Also removed a prefix for many of the BlueprintNames that were .. "test_data/complex". 
This is not how we want to specify BlueprintNames, like paths.


ingeridhellen (approved):
Nice cleanup! 🎉

PR #568, opened 2023-09-15
fix: validate blueprint attribute default

### LINKED ISSUE #567: fix: validate default values from blueprints

collinlokken:
Validates defaults defined in blueprints according to the attribute type

Currently defaults are not checked and invalid defaults can be defined

Closed #567

### DIFF
--- a/src/common/utils/validators.py
+++ b/src/common/utils/validators.py
@@ -146,7 +146,32 @@ def _validate_entity(
                 attributeDefinition, entity[attributeDefinition.name], f"{key}.{attributeDefinition.name}"
             )
         elif attributeDefinition.attribute_type == "any" and attributeDefinition.name == "default":
-            continue
+            default_attribute_definition = BlueprintAttribute(
+                name=attributeDefinition.name,
+                attribute_type=entity["attributeType"],
+                dimensions=entity.get("dimensions", None),
+            )
+            if default_attribute_definition.is_array:
+                _validate_list(
+                    entity["default"],
+                    default_attribute_definition,
+                    get_blueprint,
+                    f"{key}.{default_attribute_definition.name}",
+                    len(default_attribute_definition.dimensions.dimensions),
+                    "extend",
+                )
+            elif default_attribute_definition.is_primitive:
+                _validate_primitive_attribute(
+                    default_attribute_definition, entity["default"], f"{key}.{default_attribute_definition.name}"
+                )
+            else:
+                _validate_complex_attribute(
+                    default_attribute_definition,
+                    entity["default"],
+                    get_blueprint,
+                    f"{key}.{default_attribute_definition.name}",
+                    "extend",
+                )
         else:
             _validate_complex_attribute(
                 attributeDefinition,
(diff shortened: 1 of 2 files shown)


src/common/utils/validators.py
```diff
@@ -146,7 +146,28 @@ def _validate_entity(
...
+                dimensions=entity.get("dimensions", None),
+            )
+            if default_attribute_definition.is_array:
+                _validate_list(
+                    entity["default"],
+                    default_attribute_definition,
+                    get_blueprint,
+                    f"{key}.{default_attribute_definition.name}",
+                    len(default_attribute_definition.dimensions.dimensions),
+                    implementation_mode,
```
ingeridhellen, line 161: We want to check that the default value is valid in itself (ie it implements its own type) and that it is either of same type as defined in the attribute definition or extending from it
```suggestion
                    "extend",
```

src/tests/bdd/entity/validate_default_entity.feature
```diff
@@ -0,0 +1,117 @@
...
+    """
+    Then the response status should be "OK"
+
+    Given i access the resource url "/api/entity/validate-existing-entity/data-source-name/$Erna"
+    When i make a "POST" request
+    Then the response status should be "OK"
+    And the response should contain
+    """
+    "OK"
+    """
```
ingeridhellen, line 113: I don't see why you need these tests. You want to test the default validation, and to do that you only need to test the blueprint

src/tests/bdd/entity/validate_default_entity.feature
```diff
@@ -0,0 +1,117 @@
...
+          "default": "Norwegian"
+        },
+        {
+          "name": "prime_minister",
+          "type": "dmss://system/SIMOS/BlueprintAttribute",
+          "attributeType": "dmss://data-source-name/root_package/PersonBlueprint",
+          "label": "Prime Minister",
+          "optional": true,
+          "default": {
+            "type": "dmss://data-source-name/root_package/PersonBlueprint",
```
ingeridhellen, line 70: This is nice enough, but I would also like to test a default value where default.type extends from attributeType

src/tests/bdd/entity/validate_default_entity.feature
```diff
@@ -0,0 +1,117 @@
...
+            "name": "Jonas",
+            "nationality": "Norwegian"
+          }
+        },
+        {
+          "name": "lngLat",
```
ingeridhellen, line 76: Could we have an array of complex values too?

src/tests/bdd/entity/validate_default_entity.feature
ingeridhellen, line 1: We should test an incorrect blueprint too. Ie with invalid defaults

ingeridhellen (changes requested):
Nice work digging though DMSS - that is no easy task! I still think there remains some work here, though, especially on the test coverage

src/tests/bdd/entity/validate_default_entity.feature
collinlokken: Agreed

src/tests/bdd/entity/validate_default_entity.feature
collinlokken: No, agreed, it does not make sense to test this here. (removed)

src/tests/bdd/entity/validate_default_entity.feature
collinlokken: Added a couple of optional complex parents 👪

ingeridhellen (approved):
Much better! I still think we could have more invalid examples, though, but this is a good start

Oh, and please don't force push during a review - it makes it so much harder for the reviewer to see what has changes since last time. Just make a regular commit, then squash it once the PR is approved

Issue #495, opened 2023-08-09
refactor(test): UNIT TEST REFACTOR  EPIC


KristianKjerstad:
# Compelete Unit test Refactor

✅   Unit tests files should start with "test_"

✅   Split up the mocking class so that it is easier to add/remove stuff in the future

✅   Moved all blueprints into `mock_blueprints` folder, with one `.json` for each

✅   Refactored `MockBlueprintProvider` for top-level unit tests, which reads from the `mock_blueprints_folder`

✅   Refactored all the unit tests under `test-data-structure/`

> ✅  Test dot notation
> ✅  Test Find
> ✅  Test GroupBy
> ✅  Test HasKeyValuePairs
> ✅  Test Traverse

✅   First Iteration of refactor the tests under the folder `test_tree_functionality/` (in review)

 > ✅ Moved all blueprints under `mock_blueprints_for_tree_tests/`
 > ✅ Made a `MockBluePrintProvider` for tree tests
 > ✅ Cut all references outside the tree test folder
 > ✅ Made `Box`, `Chest`, `Room`- example

✅  Simpler `MockBlueprintProvider`, reads type and looks up type.blueprint.json

✅ Make it explicit what SIMOS blueprints are reffered to in the test. 

✅ Split up tree_node_convertion() into the two cases. 

✅ Split up test_tree_node_to_ref dict and the other one

✅ Move all tests into folders so that it is equal to the `src/` folder view

✅ Make the mocking of ´BlueprintProvider´ be data-less

> ✅ Made SIMOS blueprints required to be hard-coded from the test themselves, to make it clear what blueprints are used in the tests
> ✅ Made the mocking happen only in the setUp methods of the tests to make them more clean
> ✅ Make it required to put in the path, 
> ✅ Make it required to reference all the blueprints you want to use in the test. 
> ✅ Move blueprints "down" to each test and update the reference

✅  Create Default Array - Don't use local blueprints
>   ✅ also use the SetUP method, like refactor this whole test

✅ Skille ut tree tests slik sånn at den blir lik de andre. 

✅ Don't use Get and print diff instead use normal ... differ

✅ Go through tests and use `entity = {}` and `deepcopy( )` pattern

✅   Test-document-service-get-document should use the correct mocker

✅  Split the mock recipes into separate files, like the blueprints are done. 

✅   Bytte til at mockers bruker riktig get_blueprint (with extensions)
✅  Verifisere at dette funker ved å la blueprints extende [NamedEntity]

PR #575, opened 2023-09-19
test: made simos blueprints explicit

pbullhove:

In order to make it very clear what CORE: blueprints are accessed in the tests, the `MockBlueprintProvider( )` class is modified to take a list of available simos blueprints as a parameter. 

This is one step on the way to making the mocker be data-less, taking all data as parameters. 
The end goal is that the mocker does not know any data, only gets data as parameters, so the tests themselves provide the data explicitly. 

Many tests are now modified to facilitate this change

#495


eoaksnes (approved):
Good work! :)

PR #576, opened 2023-09-19
test: made simos blueprints explicit

pbullhove:

The mocking objects should only be created in the setUp methods. 
Then in the tests, they are modified instead of created. 

Concistency between tests. 
Makes the tests more clean, ie. readable

#495

### DIFF
--- a/src/tests/unit/common/utils/validators/test_validate_entity.py
+++ b/src/tests/unit/common/utils/validators/test_validate_entity.py
@@ -9,14 +9,14 @@ from tests.unit.mock_data.mock_recipe_provider import mock_storage_recipe_provid
 
 
 class ValidateEntityTestCase(unittest.TestCase):
-    simos_blueprints = [
-        "dmss://system/SIMOS/Blueprint",
-        "dmss://system/SIMOS/NamedEntity",
-        "dmss://system/SIMOS/BlueprintAttribute",
-    ]
-    mock_blueprint_provider = MockBlueprintProvider(simos_blueprints_available_for_test=simos_blueprints)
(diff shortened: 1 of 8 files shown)


src/tests/unit/common/utils/validators/test_validate_entity.py
```diff
@@ -9,14 +9,14 @@
...
-    mock_blueprint_provider = MockBlueprintProvider(simos_blueprints_available_for_test=simos_blueprints)
-    mock_document_service = get_mock_document_service(blueprint_provider=mock_blueprint_provider)
-    get_blueprint = mock_document_service.get_blueprint
+    def setUp(self) -> None:
+        simos_blueprints = [
+            "dmss://system/SIMOS/Blueprint",
+            "dmss://system/SIMOS/NamedEntity",
+            "dmss://system/SIMOS/BlueprintAttribute",
+        ]
+        self.mock_blueprint_provider = MockBlueprintProvider(simos_blueprints_available_for_test=simos_blueprints)
```
eoaksnes, line 18: Not 100% overview, but does it make sense if we always "go through" the document service to call get_blueprint (that uses the blueprint provider), does it make sense to consider taking in only the simos_blueprints to the get_mock_document_service, and move the creation of the mock_blueprint_provider inside that method?

```python 
def get_mock_document_service(
    blueprint_provider,
    repository_provider=None,
    recipe_provider=mock_storage_recipe_provider,
    user=None,
    context=None,
):
    return DocumentService(
        blueprint_provider=blueprint_provider,
        recipe_provider=recipe_provider,
        repository_provider=repository_provider,
        user=user,
        context=context,
    )
```

eoaksnes (approved):

PR #580, opened 2023-09-24
feat: populate data source at startup using environment variable

eoaksnes:

* Populate data source from environment variable called `DATA_SOURCE_FILES` (maybe there will be more than one data source in the future)
* Reset the data source at startup

### DIFF
--- a/src/init.sh
+++ b/src/init.sh
@@ -1,7 +1,6 @@
 #! /bin/bash
 set -eu
 
-
 echo "########### VERSION ##########"
 if [[ -e "/code/src/version.txt" ]]; then
   cat "/code/src/version.txt"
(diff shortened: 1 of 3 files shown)


src/init.sh
```diff
@@ -10,8 +9,15 @@ else
 fi
 echo -e "########### VERSION ##########\n"
 
-
 if [ "$1" = 'api' ]; then
+  if [ "${DATA_SOURCE_FILES:-""}" != "" ]; then
+    echo "$DATA_SOURCE_FILES" > /code/src/home/system/data_sources/system.json
+  fi
+
+  if [ "${RESET_DATA_SOURCE:-"on"}" == "on" ]; then
```
collinlokken, line 17: blir det litt rart å sjekke for "on"?  kanskje heller true/false?

collinlokken (approved):
Nice!

PR #581, opened 2023-09-25
feat: support app registration with a Federated Credential to login

eoaksnes:

* Allow the Application to log in as itself (in Github action)


ingeridhellen (approved):
Can't say I understand this, though

PR #583, opened 2023-09-25
test: removed duplicate methods testing the same thing. 

pbullhove:

Found two testing methods that seem to be the exact same test case. 
Removed one of them. Renamed the method name of the other 

Want the reviewer to double check that this is correct.


ingeridhellen (approved):
I like the new names!

PR #587, opened 2023-09-26
test: mock storage recipe provicer data-less

pbullhove:

This PR is the first of multiple in the quest to make the recipe-provider data-less. 
This first step moves all the storage recipes out from the mocker, instead reading it from file.


collinlokken (approved):
Nice that the test data is moved out of the test itself :)

PR #588, opened 2023-09-26
test: 👩‍🍳 More on the mock-recipe-provider

pbullhove:

This is the next PR in line for making the mock-recipe-provider data less. 
So we want each tests to be able to specify the locations for the storage recipes, so that the storage recipes can be moved 'down' to the test. 

1. Created a class, which returns a mocker with the path specified. 
2. Removed calls to the mocker where it was not required.

### DIFF
--- a/src/tests/unit/mock_data/mock_recipe_provider.py
+++ b/src/tests/unit/mock_data/mock_recipe_provider.py
@@ -5,20 +5,23 @@ from domain_classes.storage_recipe import StorageAttribute, StorageRecipe
 from enums import StorageDataTypes
 
 
-def mock_storage_recipe_provider(type: str, context: str) -> list[StorageRecipe]:
-    all_storage_recipes = defaultdict(list)
-    with open("src/tests/unit/mock_data/mock_storage_recipes/mock_storage_recipes.json") as f:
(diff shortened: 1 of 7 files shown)


src/tests/unit/mock_data/mock_recipe_provider.py
```diff
@@ -5,20 +5,23 @@
...
-            attributes={
-                a["name"]: StorageAttribute(
-                    name=a["name"], contained=a["contained"], storage_affinity=a.get("storageAffinity", "default")
-                )
-                for a in sr.get("attributes", [])
-            },
-        )
-        for sr in all_storage_recipes[type]
-    ]
+def mock_storage_recipe_provider_generator(path_to_mock_storage_recipes: str = None):
```
collinlokken, line 8: Interesting. Maybe this should be a class, and the `path` parameter could be a data field? Might get around this function-inside-function situation, which I find a little confusing. Would love to discuss this further.

collinlokken: recipe_provider would then resemble blueprint_provider, which makes sense

collinlokken (approved):
Nice! The use of class in the recipe provider makes it more understandable 😄

PR #593, opened 2023-09-27
refactor: for separation of concern: moving 'default' and 'constructor' into the classes

pbullhove:

So the default `user` and the default `AccessControlList` were not together with the class, which made it quite messy. 
Now there is separation of concern. 

1. Removed the constant `DEFAULT_ACCESS_CONTROL_LIST`, and moved that into classmethod `AccessControlList.default()`
2. Moved the ´create_access_control_list(user)´ method into the classmethod ´AccessControlList.create_with_owner()´ which is what it actually is used for. 
3. Removed the config `default_user`, and moved that into the classmethod `User.default()`
4. Updated usages for `default_user` and `DEFAULT_ACCESS_CONTROL_LIST`.


ingeridhellen (approved):
Nice!

PR #609, opened 2023-09-29
feat: roles attribute on view config

collinlokken:
Adds "roles" attribute to ViewConfig attribute type 

Make role based filtering possible in the frontend


pbullhove (approved):
Nice

PR #610, opened 2023-09-29
Refactor: Get Role Assignments for separation of concern and Testability

pbullhove:

This is part of a series of PRs that are refactoring `authentication`, due to it being hard to test. 
A lot of that were in wrong files, functions did too many things etc. 
Now this tries to simplify things and make it easier to test

1. Rename a file to `authentication/utils.py`
2. add `is_expired()` helper method to PAT
3. Add class `RoleAssignmentsProvider`, and use that as input parameter to a method in `utils` so that it can be tested
4. Move an exception instance, `credentials_exception` into the file where it is used


collinlokken (approved):
Nice!

PR #611, opened 2023-10-02
Refactor: Move api client into services instead of utils

pbullhove:

IMO `utils` should be utility functions, not API clients. That should be in the services folder

SO this PR moves an API client into `services`


collinlokken (approved):
Nice

PR #613, opened 2023-10-02
Fix: Not throw on empty role assignments

pbullhove:

There is a method which was throwing since it used direct access with [], now switched to `dict.get()` so it does not throw

Also changed `.warn()` -> `.warning()` since `.warn()` is deprecated


eoaksnes (approved):
Nice :)

PR #619, opened 2023-10-02
fix: adjust depth to avoid resolve more than needed

eoaksnes:

* Avoid resolving more than needed


pbullhove (approved):
Approved, but I don't really understand why we will never need more than 2 here

eoaksnes:
> Approved, but I don't really understand why we will never need more than 2 here

It is because we check if a name (package) already exists before adding things to a package. The content is not contained in a package, so we need to have depth 2 to resolve the name, but In theory, we can adjust to 2, and later in the code just resolve the references before checking identical package names.

PR #620, opened 2023-10-02
refactor: remove unused methods in tree node

pbullhove:

So there was like 7-8 methods in tree node that had 0 usages. 
That class is quite large and hard-to-read, so it's nice to clean it up a bit. 

The methods were mostly small helper methods, which can be easily re-written or found in the git log if they are required again 

Because we need to keep it simple, small, and easy to read


eoaksnes (approved):
Awesome! Very nice! 

image

PR #623, opened 2023-10-03
Test: Start of tree-node unit tests refactor

pbullhove:

We created a blueprint mocker a while ago that was to be used in all the unit tests. 
Now we are starting to use that in the `tree_tests` as well. 

Also some refactoring in the tests to have less duplication of code. More happens in the setup.


soofstad (approved):
:+1:

PR #636, opened 2023-10-04
fix: remove item from required list attribute

### LINKED ISSUE #632: investigate: can we delete a single item inside a list (without sending the whole list)?

soofstad:

- Will now not raise exception if removing an item from a required list
- "content" in Package.json is not required (makes sense, and useful for testing)
- Added a BDD test for this scenario 

It should be allowed to delete items from a list, even if the list attribute itself is required.

closes #632

### DIFF
--- a/src/services/document_service.py
+++ b/src/services/document_service.py
@@ -227,7 +227,7 @@ class DocumentService:
 
     def remove(self, address: Address) -> None:
         node = self.get_document(address)
-        if node.parent and not node.attribute.is_optional:
+        if node.parent and not node.is_optional:
             raise ValidationException("Tried to remove a required attribute")
 
         data_source = self.repository_provider(address.data_source, self.user)
--- a/src/common/tree/tree_node.py
+++ b/src/common/tree/tree_node.py
@@ -47,6 +47,14 @@ class NodeBase:
         self._type = value
 
     @property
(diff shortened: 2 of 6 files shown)


src/services/document_service.py
```diff
@@ -242,7 +242,7 @@ def get_document(self, address: Address, depth: int = 0) -> Node | ListNode:
 
     def remove(self, address: Address) -> None:
         node = self.get_document(address)
-        if node.parent and not node.attribute.is_optional:
+        if node.parent and not node.attribute.is_optional and not node.parent.is_array():
```
pbullhove, line 245: Question: 
What if there is a required element in array? Would that not pass through this test? 

Do we not want to check if the element itself in the array is optional, not whether the parent array is optional?

pbullhove: Do the children of array inherit the 'optional' status of parent array?

pbullhove (approved):
Approved, but with some questions.

src/services/document_service.py
soofstad: We have no concept to say that items in a list is required or not.

Children of arrays to inherit the "node.attribute" instance. Added a "is_optional" to NodeBase to make this check a bit more intuitive.

PR #640, opened 2023-10-05
refactor: move into providers folder

pbullhove:

Right now there is much unrelated code in the `utils` folder which makes it hard to navigate. 

There is a software architecture principle
- Group related code

Also order to have a more screaming architecture
- Have folder names which 'scream' what they are doing. 

Therefore this PR. 
1. move `remove_document.py` which is a helper method for the document service into a folder under services called `document_service`
2. create a folder called `common/providers`, and group `blueprint_provider`, `address_resolver` etc in this folder


eoaksnes (approved):
👍

PR #641, opened 2023-10-05
refactor: Delete unused method + Move helpers into 'entity'

pbullhove:

1. Was a method `to_dot_notation` that was never used. Deleted together with it's tests. 
2. Moved some entity helper methods into `common/entity`


soofstad (approved):
Don't see any helper methods that has been moved

PR #646, opened 2023-10-06
refactor: add parameters to blueprint_provider so that it can be tested + Delete unused code

pbullhove:

So in order to make the blueprint provider testable, we add the dependencies parameters to the method. 
-> Dependency injection 

Also delete some code in `address_resolver` that was not used from anywhere


eoaksnes (approved):
Smooth 👍

PR #655, opened 2023-10-09
Test Blueprints now inherit from NamedEntity

pbullhove:

In very many cases the blueprints don't inherit from NamedEntity, and by doing this they can be reduced in size. 

This PR concludes the unit test refactor epic


collinlokken (approved):
Nice 👍

PR #662, opened 2023-10-11
Refactor: A bit of address resolver

pbullhove:

So instead of the document service doing a little bit of address resolving work, some of the logic was moved into the address  resolver. Namely the responsability for resolving entities with depth flag. 

This keeps the responsabilities more separated. ie. clean code

This also made the `reference_resolver.py` have no extenral dependencies outside of address_resolver, so it was moved into there, to keep coupled things close together


soofstad (approved):

eoaksnes (approved):
👍

PR #671, opened 2023-10-18
fix(export): add proper error response on export ++ minor issues

soofstad:
Trying to export "system/SIMOS" revealed some missing error responses.
When that was added, some invalid blueprints and entities where discovered.

These has now been fixed.


collinlokken (approved):
Nice!

PR #672, opened 2023-10-26
fix: transform entities into references when storage uncontained

collinlokken:
Changes the behaviour of document save by creating a (storage) reference to entities only if the entity is storage uncontained.


soofstad (approved):
Look great!

PR #673, opened 2023-10-27
feat: get_attribute able to resolve references

soofstad:

- GetAttribute endpoint will now resolve references by default, and return "direct address"

Some plugins create views of children that are references. Should be possible to see the actual content.

### DIFF
--- a/src/features/attribute/use_cases/get_attribute_use_case.py
+++ b/src/features/attribute/use_cases/get_attribute_use_case.py
@@ -1,16 +1,22 @@
 from authentication.models import User
 from common.address import Address
 from common.tree.tree_node import Node
+from enums import SIMOS
 from services.document_service.document_service import DocumentService
 from storage.internal.data_source_repository import get_data_source
 
 
 def get_attribute_use_case(
     address: str,
+    resolve: bool,
     user: User,
     repository_provider=get_data_source,
 ):
     """Get attribute by reference."""
     document_service = DocumentService(repository_provider=repository_provider, user=user)
     node: Node = document_service.get_document(Address.from_absolute(address))
-    return node.attribute.to_dict()
+    direct_address = address
+    if resolve and node.attribute.attribute_type == SIMOS.REFERENCE.value:
+        direct_address = node.entity["address"]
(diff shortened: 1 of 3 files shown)


src/features/attribute/use_cases/get_attribute_use_case.py
```diff
@@ -1,16 +1,23 @@
...
     user: User,
     repository_provider=get_data_source,
 ):
     """Get attribute by reference."""
     document_service = DocumentService(repository_provider=repository_provider, user=user)
     node: Node = document_service.get_document(Address.from_absolute(address))
-    return node.attribute.to_dict()
+    direct_address = address
+    if resolve:
+        if node.attribute.attribute_type == SIMOS.REFERENCE.value:
```
collinlokken, line 20: maybe `and` here instead of `if`x2?

src/features/attribute/use_cases/get_attribute_use_case.py
```diff
@@ -1,16 +1,23 @@
...
+            direct_address = node.entity["address"]
+            node: Node = document_service.get_document(Address.from_absolute(direct_address), 2)  # type: ignore
+    return {"attribute": node.attribute.to_dict(), "address": direct_address}
```
collinlokken, line 23: Does it make sense to send the input (address) back? Wouldn't client already have this value?

collinlokken (changes requested):
Nice, but maybe the `direct_address`should be processed further before we return it?

src/features/attribute/use_cases/get_attribute_use_case.py
soofstad: yepp. Fixed

src/features/attribute/use_cases/get_attribute_use_case.py
soofstad: The client does not. It only has the "scope", eg. ".a.b".
The address returned here is the value of the "address" in the reference, if it's a reference.

collinlokken (approved):
Nice!

src/features/attribute/use_cases/get_attribute_use_case.py
collinlokken: You are right!

PR #676, opened 2023-10-27
fix: defer checking of duplicate name in empty package references

collinlokken:
Fix that skips checking of duplicate name in empty references

If references are attempted created before the document itself is created, then we assume that you are an expert and know what you are doing.

### DIFF
--- a/src/services/document_service/document_service.py
+++ b/src/services/document_service/document_service.py
@@ -143,22 +143,8 @@ class DocumentService:
         # If the node is a package, we build the path string to be used by filesystem like repositories.
         # Also, check for duplicate names in the package.
         if node.type == SIMOS.PACKAGE.value:
-            if len(node.children) > 0:
-                contentListNames = []
-                for child in node.children[0].children:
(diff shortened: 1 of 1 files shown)


src/services/document_service/document_service.py
```diff
@@ -269,3 +255,28 @@ def remove(self, address: Address) -> None:
...
+                if child.type == SIMOS.REFERENCE.value:
+                    try:
+                        entity = self.get_document(Address(child.entity["address"], data_source_id), depth=0).entity
+                    except NotFoundException:
+                        """
+                        reference does not point to anything because the document it points to has not been uploaded
+                        yet. dm-cli was probably used
+                        """
+                        logger.warning("An empty reference was received. I hope you know what you are doing...")
+                        entity = {}
```
soofstad, line 272: Looks like we can just do a "continue" here. Skipping the rest of the loop

soofstad (approved):
Approved if comment is resolved :slightly_smiling_face:

PR #683, opened 2023-11-01
fix: smarter validate attributes of type any

soofstad:
- A bit more robust handling of "any" type and validation
- "any" is now considered a primitive type, which was needed for being able to update primitive values
- Reorder a list of "if/elif" in validators.py, so that the special case of BlueprintAttributes get's picked out before "any" is picked as a primitive
 
- The validator did not allow users to update the "default" attribute in BlueprintAttribute with a primitive value

none


eoaksnes (approved):
👍  good

PR #688, opened 2023-11-07
refactor: move reference_resolver into address_resolver

pbullhove:

Similar as an older PR that was approved a month ago, but not merged. 

This moved a the reference resolver into the address resolver and refactors it to be much more readable.


soofstad (approved):
Good change!

PR #691, opened 2023-11-07
fix(node): Node.node_id() return invalid address to list elements

soofstad:
- Have Node.node_id() return a valid address for items in a list

When using add_document(), I expect to be able to use the returned address to fetch the newly created entity.

blocking "recurring job handler"


eoaksnes (approved):
👍 Looks good!

PR #694, opened 2023-11-10
Fix/create entity use default on optional

### LINKED ISSUE #684: feat: create_entity of type blueprint should insert default attributes
- [ ] when creating a new blueprint, type, name, description  should be automatically added

soofstad:
NOTE: Depends on #693 

 Fixes a bug in CreateEntity where optional attributes were not instantiated when provided default values.

Was not behaving as intended

closes #684


collinlokken (approved):
Looks good 👍

PR #697, opened 2023-11-14
fix: address_resolver did not use new datasource name

soofstad:
- Fix a bug in address_resolver where it did not use the new data_source when resolving a reference to another data_source
- Add test that catches this case

bug
none


collinlokken (approved):
Looks good 👍

PR #701, opened 2023-11-17
build: remove publish pypi package

### LINKED ISSUE #681: ci(dmss): publish DMSS API package to PyPI fails
Just turn off publish in the workflow. Maybe not delete the code, but just turn it off.

soofstad:
No need for it, and it fails to run.
closes #681


pbullhove (approved):
Nice. 
One can also do this in gituhub actions, by clicking "disable". 
But this works as well

PR #720, opened 2023-12-06
fix: get attributes for address that points to reference

henrikbossart:
Fixes a bug where its not possible to resolve "chained" references.

### DIFF
--- a/src/features/attribute/use_cases/get_attribute_use_case.py
+++ b/src/features/attribute/use_cases/get_attribute_use_case.py
@@ -18,5 +18,9 @@ def get_attribute_use_case(
     direct_address = address
     if resolve and node.attribute.attribute_type == SIMOS.REFERENCE.value:
         direct_address = node.entity["address"]
-        node: Node = document_service.get_document(Address.from_absolute(direct_address), 2)  # type: ignore
+        # type: ignore
+        reference_node: Node = document_service.get_document(
+            Address.from_relative(direct_address, None, Address.from_absolute(address).data_source), 2
+        )
+        return {"attribute": reference_node.attribute.to_dict(), "address": direct_address}
     return {"attribute": node.attribute.to_dict(), "address": direct_address}


src/features/attribute/use_cases/get_attribute_use_case.py
```diff
@@ -18,5 +18,9 @@ def get_attribute_use_case(
     direct_address = address
     if resolve and node.attribute.attribute_type == SIMOS.REFERENCE.value:
         direct_address = node.entity["address"]
-        node: Node = document_service.get_document(Address.from_absolute(direct_address), 2)  # type: ignore
+        # type: ignore
+        reference_node: Node = document_service.get_document(
+            Address.from_relative(direct_address, None, Address.from_absolute(address).data_source), 2
```
soofstad, line 23: Could we add a few comments to this?
I'm really struggling to see what is going on :smiling_face_with_tear:

soofstad (changes requested):

src/features/attribute/use_cases/get_attribute_use_case.py
soofstad: Like what are the different addresses, why are we calling "from_relative()" on `direct_address` (that _sounds_ wrong, even if it probably isn't

henrikbossart: Comments would probably be nice, yes.
I'm honestly not quite sure whats going on here myself, so maybe @eoaksnes can chime in and explain a bit better, but in essence;

When referencing an entity that references a file, you somehow lose the datasource somewhere... No idea why...

We use the `from_relative`, because that one uses the `from_absolute` if the `direct_address` contains a data source..

Also, doing this fixed the issue babak was having with images (I hope), and we were struggling with this for quite some time 😅

henrikbossart: No, I think the issue is that its not a part of the file id then... 🤔 We can look at this together tomorrow 👍

henrikbossart: @eoaksnes tagging you here since you have more insight into this than I do :+1:

PR #724, opened 2023-12-14
feat: relative references

eoaksnes:

* Feat: Be able to use `~´  inside an address to point/go upwards to attributes.

### DIFF
--- a/src/common/address.py
+++ b/src/common/address.py
@@ -36,10 +36,24 @@ class Address:
         return cls(path, address, protocol)
 
     @classmethod
-    def from_relative(cls, address: str, document_id: str | None, data_source: str):
+    def from_relative(
+        cls, address: str, document_id: str | None, data_source: str, relative_path: list[str] | None = None
+    ):
         if "://" in address:
             # Already contains protocol and data source
             return cls.from_absolute(address)
+        if address.startswith("~"):
+            if not relative_path:
+                raise ApplicationException(
+                    "Missing the relative path and therefore it is not possible to resolve the ~ inside the address."
+                )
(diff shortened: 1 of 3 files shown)


src/common/address.py
```diff
@@ -36,10 +36,24 @@ def from_absolute(cls, address: str):
...
+        if address.startswith("~"):
+            if not relative_path:
+                raise ApplicationException(
+                    "Missing the relative path and therefore it is not possible to resolve the ~ inside the address."
+                )
+            go_up = address.count("~")
+            path = relative_path.copy()
+            while go_up != 0:
+                path.pop(0)
+                go_up -= 1
```
collinlokken, line 54: Another way could be to replace ~with ../ and use os.path.normpath to normalize it

collinlokken (approved):
👍

src/common/address.py
```diff
@@ -36,10 +36,24 @@ def from_absolute(cls, address: str):
...
         if "://" in address:
             # Already contains protocol and data source
             return cls.from_absolute(address)
+        if address.startswith("~"):
+            if not relative_path:
+                raise ApplicationException(
+                    "Missing the relative path and therefore it is not possible to resolve the ~ inside the address."
+                )
+            go_up = address.count("~")
+            path = relative_path.copy()
```
soofstad, line 51: Might be good enough for now. We there should at least be a check on unsupported use of `"~"`.
relative references like ".data.~.data" will not work now. Which probably is fine.

Can just check that if there is any number of "~.", they must all be at the start of the reference

soofstad (approved):

PR #726, opened 2023-12-16
fix: resolve tilde correct

eoaksnes:

* Bug in how the resolve tilde ~ (the fix is to remove the last element in the path, not the first)
* Include the index to the relative path if used inside a list
* Raise exception if going outside document

### DIFF
--- a/src/common/address.py
+++ b/src/common/address.py
@@ -50,9 +50,17 @@ class Address:
             go_up = address.count("~")
             path = relative_path.copy()
             while go_up != 0:
-                path.pop(0)
+                if len(path) == 0:
+                    raise ApplicationException(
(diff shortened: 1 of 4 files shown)


src/common/address.py
```diff
@@ -50,9 +50,15 @@ def from_relative(
             go_up = address.count("~")
             path = relative_path.copy()
             while go_up != 0:
-                path.pop(0)
+                if len(path) == 0:
+                    raise ApplicationException("Trying to go outside document.")
```
soofstad, line 54: This message could be more specific?
For example; 
"Invalid relative reference. Traversing outside a contained document with '~' is not supported"

soofstad (approved):

PR #729, opened 2023-12-18
fix: delete items in list

eoaksnes:

* fix: delete item in a list fails due to find method wants [0] and not only the index position


l3abak (approved):
ok

PR #766, opened 2024-01-24
refactor: only resolve relevant references

eoaksnes:

* Improve performance by only resolve relevant references

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,8 +16,8 @@ services:
     depends_on:
       - db
   db:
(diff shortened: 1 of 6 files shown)


soofstad (approved):
Good work :+1: 
If there are some ideas how this code could be improved, could add them in some `# TODO:` 's

pbullhove (approved):
I agree. 
Looks ok, but I think the main thing here is that the whole address_resolver could get some love to make more easily readable. 
But that's not a priority now!

soofstad (approved):

PR #776, opened 2024-02-06
perf: blueprint cache shared between users

soofstad:
- All users now share the same blueprint cache

- Caching was not working as intended. With a new cache for each user and each instance of the DocumentService() (which is created on each request)

related to #770


pbullhove (approved):
seems sketchy but ok

PR #780, opened 2024-02-09
Fix: cache blueprint provider instead of one instance of it

pbullhove:

Earlier we instantiated one. 

Think this is better to be able to reintroduce security measures. 
This is better because now we can have cache and userchecking inside the blueprint provider later I think


soofstad (commented):
Not sure I see the benefit of separating out the functions from the class. The functions are not used anywhere else, and only lead to more nested code.

This is likely to be reworked anyhow when we refactor the storage system.

PR #781, opened 2024-02-12
refactor: keep content type when uploading file

awesthouse:
Saves content-type as metadata when uploading file.

Right now we are using the filetype (extension) and a static list of all mimetypes to map the  mimetype in frontend. Would be better to keep the content-type on upload.

### QUESTION:  
Does setting contentType to File Blueprint create some problems for any existing files? It's not optional


soofstad (commented):
I think this test is misleading. It's the only place we are actually sending content-type.

If we want to require all clients to send a "correct" contentType, not just "octet-stream", that will require changes to all clients. 
and in my opinion, does not really solve anything.
We only move the mimetype-database from our UiPlugin, to every client who wants to upload.

PR #783, opened 2024-02-13
feat: add azure app insight

eoaksnes:

* Send logs to Azure app insights


src/common/responses.py
```diff
@@ -98,7 +99,7 @@ async def wrapper_decorator(*args, **kwargs) -> TResponse | JSONResponse:
             except ValidationException as e:
                 if logger.level <= logging.DEBUG:
                     traceback.print_exc()
-                logger.error(e)
+                logger.error(e, extra={"Traceback": get_traceback()})
```
jorgenengelsen, line 102: Consider if this can be skipped.  Traceback will log user-input, and ValidationException "should" be easy to provide the user with a readable error message?

jorgenengelsen (commented):
Looks good to me! Left some comments to look at.

src/common/responses.py
```diff
@@ -108,7 +109,7 @@ async def wrapper_decorator(*args, **kwargs) -> TResponse | JSONResponse:
             except BadRequestException as e:
                 if logger.level <= logging.DEBUG:
                     traceback.print_exc()
-                logger.error(e)
+                logger.error(e, extra={"Traceback": get_traceback()})
```
soofstad, line 112: IMO, these are not errors. They are discovered, handled, and known, execution aborts based on "BadInput". If you still want to see that caused that, it should be logged as "DEBUG". 

Same goes for the rest of the changes in this file

soofstad (changes requested):

src/common/responses.py
eoaksnes: 👍  the plan is to go through all log statements after we have started logging to see what we would like to log, so seperate PR

src/common/responses.py
```diff
@@ -99,32 +100,36 @@ async def wrapper_decorator(*args, **kwargs) -> TResponse | JSONResponse:
...
                     traceback.print_exc()
-                logger.error(e)
+                logger.debug(e)
                 return JSONResponse(e.dict(), status_code=status.HTTP_404_NOT_FOUND)
             except BadRequestException as e:
                 if logger.level <= logging.DEBUG:
                     traceback.print_exc()
-                logger.error(e, extra={"Traceback": get_traceback()})
-                logger.error(e.dict())
+                logger.debug(e, extra={"Traceback": get_traceback()})
```
soofstad, line 113: Is the error stack now printed twice? Both on line 112 and 113? Or is extra= not visible in console?

soofstad (approved):

PR #789, opened 2024-02-15
feat: option to enforce uid on all complex children

### LINKED ISSUE #785: Unique ids in tables and list
* Add _id to Entity (optional)
* Add access_by_id (or something) to BlueprintAttribute
* POST documents/ should add ID if access_by_id is true
* PUT documents/ should also handle ID if access_by_id is true

soofstad:
- New BlueprintAttribute - attribute "ensureUID"
  - This means that any complex attribute, where the attribute definition has this set to "TRUE", MUST have an "_id" key.
  - The "add_document" endpoint will inject this id if missing
- Fix some bdd tests to not use the "post file" fixture when not posting a file (should speed up tests)

- An experiment to get static identification for elements in a list

closes #785

### DIFF
--- a/src/SIMOS/BlueprintAttribute.json
+++ b/src/SIMOS/BlueprintAttribute.json
@@ -29,6 +29,14 @@
       "default": ""
     },
     {
+      "name": "ensureUID",
+      "type": "dmss://system/SIMOS/BlueprintAttribute",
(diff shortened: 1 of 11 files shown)


eoaksnes (approved):
This is good! 👍

PR #791, opened 2024-02-19
fix: gunicorn times out on some slow requests

soofstad:
- Added timit decorator for performance testing
- Increase default timout of gunicorn
- Search endpoint use the same "default_blueprint_provider"

- When running gunicorn with dm-core-packages example data, the validate request times out

related to https://github.com/equinor/dm-core-packages/issues/1307


eoaksnes (approved):
👍

PR #796, opened 2024-02-20
feat: add request profiler

eoaksnes:

* Profile any requests easily, default turned off (need to provide `profile`=true in URL to enable profiling)
* Profile requests and store to disk or Azure blob storage. Default to disk, but if `PROFILING_STORAGE_ACCOUNT` (the connection string to Azure blob storage account) is provided profiles will be uploaded to Azure blob storage.
* Can control what format to store by setting `profile_format`=html|speedscope in URL. Speedscope format is a JSON format, and html is just a simple HTML report.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -13,13 +13,13 @@ files = [
 
 [[package]]
 name = "anyio"
-version = "4.2.0"
+version = "4.3.0"
 description = "High level compatibility layer for multiple asynchronous event loop implementations"
 optional = false
 python-versions = ">=3.8"
 files = [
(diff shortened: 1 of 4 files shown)


pbullhove (approved):
Nice. Looks great

soofstad:
Is there any security measures that stops anyone from filling up the storage account with data? :thinking:

soofstad (approved):
Looks good. But might have some security issues. Filling up the disk for the container would lead to a DoS attack. Filling up the storage account could be costly

PR #811, opened 2024-03-04
fix: partial update only worked with complex

### LINKED ISSUE #810: fix: update_document partially fails on required primitives
**Bug**
update_document even with flag partial_update fails on validation when primitives are required. 

**Expected behavior**
Should be allowed to update document with only changed fields (and type attribute?) and not have to also pass required primitives in payload so that it passes validation

**Notes**
- Somewhat of a priority 'cause Babak asked for this to be fixed
- Frontend task in form is dependent on this fix (being done by @awesthouse)

soofstad:

- Fixes a bug where it was not allowed to leave out primitive values on node.update() with partial=True 
- Simplify node.update() code. No longer need to validate as much, since we do that on a later stage anyway.

closes #810

### DIFF
--- a/src/common/tree/tree_node.py
+++ b/src/common/tree/tree_node.py
@@ -255,68 +255,45 @@ class Node(NodeBase):
 
     # Replace the entire data of the node with the input dict. If it matches the blueprint...
     def update(self, data: dict, partial_update: bool = False):
+        if partial_update:
+            self.entity.update(data)
+        else:
+            self.entity = data
+
         if data.get("_id"):
             self.set_uid(data.get("_id"))
 
-        # Set self.type from posted type, and validate against parent blueprint
+        # Set self.type from posted type
         self.type = data.get("type", self.attribute.attribute_type)
-        # Modify and add for each key in posted data
-        for key in data.keys():
-            new_data = data[key]
-            if key == "_id":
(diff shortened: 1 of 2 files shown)


src/common/tree/tree_node.py
```diff
@@ -255,68 +255,45 @@ def __init__(
...
-        removed_attributes = [attr for attr in self.blueprint.attributes if attr.name not in data]
-        for attribute in removed_attributes:
-            # Pop primitive data
-            if attribute.is_primitive:
-                self.entity.pop(attribute.name, None)  # type: ignore
-            # Remove complex data
+        self.children = []
+        complex_attributes = self.blueprint.get_none_primitive_types()
+        for attribute in complex_attributes:
+            if attribute.name not in self.entity:
```
eoaksnes, line 272: Should we add a comment here? I guess this can only happen if we change the type, and then we need to remove any attribute that is not specified in the blueprint.

eoaksnes (approved):
👍 smooth

src/common/tree/tree_node.py
soofstad: This check is for optional attributes.
We now build the list of children from scratch, meaning there is nothing to remove. And it will always be correct, no matter how much you change types

PR #813, opened 2024-03-13
fix: more info on 'failed to resolve reference'

soofstad:
Add some context to the error message returned when a "resolve address" failes


pbullhove (approved):
LGTM

PR #822, opened 2024-04-05
feature: plugable repositories

soofstad:
- Change Dockerfile so that the running user also owns and installs the poetry environment (required to install additional packages at startup)
- Update README on how to use plugins
- Add import feature to Repository._get_client() method
- Update init.sh to install plugin packages at startup
- "get_blueprint" is now passed to DataSourceRepository. Needed be some repository plugins
- 

- Allows for adding repository plugins


src/storage/internal/data_source_repository.py
```diff
@@ -25,16 +24,17 @@ class DataSourceInformation(BaseModel):
...
+        self.get_blueprint = get_blueprint
 
     @staticmethod
     def validate_data_source(document: dict):
         try:
             for repo in document["repositories"].values():
                 if repo.get("name"):
                     raise BadRequestException("A repository specification may not have the 'name' key.")
-                if repo["type"] == RepositoryType.MONGO.value:
+                if repo["type"] == "mongo-db":
```
eoaksnes, line 37: Not sure why we removed the enum? :)

eoaksnes (approved):
👍 smooth

src/storage/internal/data_source_repository.py
soofstad: It's the only placed its used. So didn't see much point in having it as an enum

PR #828, opened 2024-04-17
perf(address_resolver): return early when finding a filter match

soofstad:
- Resolve one reference, check if its a match, and return. Instead of always resolving all references. Pretty big performance improvement when fetching blueprints.

- Unnecessary fetches of all references in a list of references
- Makes the first fetch of blueprints about 50% faster :smile: 

performance


eoaksnes (approved):
👍 sweet, nice work

PR #829, opened 2024-04-17
perf: cache on get_data_source

soofstad:
- DataSource caching has not been working for a while. This PR fixes this

- Reduces time for simple single document get in CosmosDB from ~1000ms to ~100ms. This is because on every request the MongoClient has to download meta data from the server before it can execute the request.


eoaksnes (approved):
👍

PR #835, opened 2024-04-29
perf: cache documents in 'resolve_document'

soofstad:
- When calling DocumentService.resolve_document(), a cache for documents are created. The cache lives for this request only.

- The cache stops DMSS fetching the same document from the data_source multiple times.
  This improved performance on some request from 82 seconds --> 18,7 seconds

perf


eoaksnes (approved):
👍 smooth

PR #851, opened 2024-09-17
Sqlclient

jta98:
Up to date sqlClient

Find function not implemented for client


src/storage/repository_plugins/sql/__init__.py
```diff
@@ -0,0 +1,299 @@
...
+        raise NotImplementedError
+
+    def delete_blob(self, uid: str):
+        raise NotImplementedError
+
+    def update_blob(self, uid: str, blob: bytearray):
+        raise NotImplementedError
+
+    def find(self, filter: dict, single=None, raw=None) -> list[dict]:
+        return []
```
soofstad, line 294: A very basic find support, `type where name == x` should be implemented.

src/system_DS_template.json
```diff
@@ -8,7 +8,7 @@
       "username": "maf",
       "password": "$MONGO_PASSWORD",
       "tls": false,
-      "database": "DMSS-core",
+      "database": "dmss",
```
soofstad, line 11: Should be no need to change this

packages/poetry.lock
soofstad, line 1: what is this "pacakges/" used for?

pyproject.toml
```diff
@@ -25,6 +25,7 @@ opentelemetry-instrumentation-fastapi = "^0.43b0"
 pyinstrument = "4.6.2"
 pydantic-settings = "^2.2.1"
 redis = {extras = ["hiredis"], version = "^5.0.4"}
+sqlalchemy = "1.4.32"
```
soofstad, line 28: Could we use ether "requirements.txt" or pyproject, not both?

src/SIMOS/SQLReference.json
soofstad, line 1: Can you explain how this is used?

soofstad (changes requested):
Nice job!
A few questions.
Also, you should run "pre-commit" in the repository to clean up the pull-request.
Please try to remove any unnecessary diff (prints, new lines, unnecessary changes etc.)

PR #864, opened 2025-03-06
refactor: increase attempts caused by cosmos db rate limiting

eoaksnes:

* Increase the number of attempts allowed for the "Request rate too large" exception, also known as error code 429, which indicates that your requests against Azure Cosmos DB are being rate limited.
* Make behave tests run again
* Make unit tests run again

* To allow the reset application script to have more attempts for adding


soofstad (approved):
Looks good!

PR #866, opened 2025-03-06
feat: add raw multiple documents and files

eoaksnes:

* Add endpoint to add multiple documents or files

PS: To fix the reset application bug I gone try to increase nginx timeout also, maybe that is enough (we dont need this pr). I am not sure if we want to bulk insert files (since can be lagre), but documents (small once) can give benefit.

* Reduce number of requests


soofstad (approved):
Good stuff!

